PR #5, opened 2020-12-10
Replace jq dependency with tinyjson.

edfloreshz:
I managed to remove the necessity for `jq` within `cargo-quickinstall`. I tried to keep it very minimal, therefore the reason of using `tinyjson`, which adds no dependencies to your crate, except for `tinyjson`, of course.


cargo-quickinstall/Cargo.toml
```diff
@@ -11,3 +11,4 @@ repository = "https://github.com/alsuren/cargo-quickinstall"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+tinyjson = "2"
```
alsuren, line 14: Please add trailing newlines here and in .gitignore (I will make a linter to catch this later)

edfloreshz: Sure thing.

cargo-quickinstall/src/main.rs
```diff
@@ -34,11 +37,22 @@ fn get_latest_version(crate_name: &str) -> std::io::Result<String> {
...
         crate_name
     );
-    bash_stdout(&command_string)
+    match bash_stdout(&command_string) {
+        Ok(stdout) => match stdout.parse() {
+            Ok(parsed) => {
+                let parsed: JsonValue = parsed;
+                let mut version: String = parsed["versions"][0]["num"].stringify().unwrap();
+                version.remove(0);
+                version.remove(version.len()-1);
```
alsuren, line 49: There might be a way to avoid needing to strip off the `"` at the start and end here.

edfloreshz: Ok, let me check!

cargo-quickinstall/src/main.rs
```diff
@@ -34,11 +37,22 @@ fn get_latest_version(crate_name: &str) -> std::io::Result<String> {
...
-            'https://crates.io/api/v1/crates/{}' \
-            | jq -r .versions[0].num",
+            'https://crates.io/api/v1/crates/{}'",
         crate_name
     );
-    bash_stdout(&command_string)
+    match bash_stdout(&command_string) {
+        Ok(stdout) => match stdout.parse() {
+            Ok(parsed) => {
+                let parsed: JsonValue = parsed;
```
alsuren, line 46: you can probably use `stdout.parse::<JsonValue>()` above, and delete this line.

I also prefer to use the ? operator so that my "happy path" code doesn't drift to the right so much.

cargo-quickinstall/src/main.rs
edfloreshz: `tinyjson`'s `stringify()` method returns the string with "" intentionally. We could create a function that does exactly the same thing but without the "". What do you think?

alsuren (approved):
Code looks like it's the right shape. I've left some minor comments, but I'm happy to merge this as-is and fix them later. I'll have another look after dinner and probably merge it. I will aim to cut a release tomorrow as well.

Thanks for submitting.

cargo-quickinstall/src/main.rs
edfloreshz: Ok, I pushed some changes using the `?` operator, I don't think `stdout.parse::<JsonValue>()` is necessary here.

cargo-quickinstall/src/main.rs
```diff
@@ -34,11 +37,18 @@ fn get_latest_version(crate_name: &str) -> std::io::Result<String> {
...
     );
-    bash_stdout(&command_string)
+    let stdout =  bash_stdout(&command_string)?;
+    let parsed: Result<JsonValue, Error> = match stdout.parse() {
+        Ok(parsed) => Ok(parsed),
+        Err(_) => Err(Error::new(ErrorKind::InvalidData, "Unable to parse JSON."))
+    };
+    let mut version: String = parsed?["versions"][0]["num"].stringify().unwrap();
+    version.remove(0);
+    version.remove(version.len()-1);
```
alsuren, line 50: Looks like JsonValue implements `TryInto<String>` https://docs.rs/tinyjson/2.2.0/tinyjson/enum.JsonValue.html#impl-TryInto%3CString%3E . It might be possible to say `let mut version: String = parsed?["versions"][0]["num"].try_into().unwrap()` and avoid the quotes.

```suggestion
    let parsed: Result<JsonValue, Error> = match stdout.parse() {
        Ok(parsed) => Ok(parsed),
        Err(_) => Err(Error::new(ErrorKind::InvalidData, "Unable to parse JSON."))
    };
    let mut version: String = parsed?["versions"][0]["num"].stringify().unwrap();
    version.remove(0);
    version.remove(version.len()-1);
```

L

alsuren: You might also be able to use .map_err() to do the error conversion. Something like:

```suggestion
    let parsed: JsonValue = match stdout.parse().map_err(|_| Err(Error::new(ErrorKind::InvalidData, "Unable to parse JSON.")))?;
    let mut version: String = parsed["versions"][0]["num"].try_into().unwrap();
```

(I've not tested this at all. It probably doesn't even compile 😁 )

My workmate has said that my error reporting "could use some work", so I'm also happy to leave your match statement as it is, and do a proper error-reporting re-work later.

PR #6, opened 2020-12-10
Replaced match with .map_err() and stringify() with try_into().

edfloreshz:
I just pushed the following changes:

1. Replaced `match` with `.map_err()` when parsing the JSON. [line 46]
2. Replaced the `.stringify()` method with `.try_into()` as you suggested, it worked. [line 47]
3. Ran `rustfmt`.


alsuren (approved):
Beautiful

PR #16, opened 2021-04-03
Improve rebuild popular crates script

sagiegurari:
Improve script based on comments in #15

### DIFF
--- a/popular-crates.txt
+++ b/popular-crates.txt
@@ -5,6 +5,8 @@
 
 cargo-quickinstall
 sensei
+ripgrep
+cargo-make
 
 b3sum
 bandwhich
@@ -56,7 +58,6 @@ cargo-lichking
 cargo-lipo
 cargo-llvm-lines
 cargo-local-registry
-cargo-make
 cargo-modules
 cargo-ndk
 cargo-outdated
--- a/rebuild-popular-crates.sh
+++ b/rebuild-popular-crates.sh
@@ -1,8 +1,11 @@
 #!/bin/bash
 
+# pup can be installed via: go get github.com/ericchiang/pup
+
+set -euo
+
 function get_top() {
-    curl https://lib.rs/$1 > ./temp.html
-    cat temp.html |
+    curl https://lib.rs/$1 |
         pup ':parent-of(:parent-of(:parent-of(.bin))) json{}' |
         jq -r '.[] |
             (.children[1].children|map(select(.class == "downloads").title)[0]// "0 ")
@@ -13,7 +16,6 @@ function get_top() {
         head -n 100 |
         sed s/^.*:// |
         sort
-    rm ./temp.html
 }
 
 top_crates=""


rebuild-popular-crates.sh
```diff
@@ -1,8 +1,11 @@
 #!/bin/bash
 
+# pup can be installed via: go get github.com/ericchiang/pup
+
+set -euo
```
alsuren, line 5:
```suggestion
set -euo pipefail
```


-o pipefail is what you need to make the pipeline fail if pup or jq fails. Otherwise the pipeline will take the exit code of `sort`, I think.

sagiegurari: that is messing up the script actually. no output. didn't investigate too much.

popular-crates.txt
```diff
@@ -110,25 +101,22 @@ critcmp
...
 exa
 fd-find
 feroxbuster
 ffsend
 flamegraph
 fw
 git-delta
-git_lab_cli
 git-state
+git_lab_cli
```
alsuren, line 119: I think that your version of sort might be ignoring - and _ characters, and only sorting alphabetic characters? I suppose we could use `sort --dictionary-order` if we want this?

sagiegurari: its a good point. can you fix that? i saw you added many more cool stuff on the way

rebuild-popular-crates.sh
```diff
@@ -1,25 +1,45 @@
...
+    (
+        cat popular-crates.txt | grep -B10000 '####################################'
+        get_top_both
+    ) | uq
 }
 
-top_crates=""
-top_crates="${top_crates} $(get_top command-line-utilities)"
-top_crates="${top_crates} $(get_top development-tools/cargo-plugins)"
-top_crates=`echo ${top_crates} | uniq | tr " " "\n" | sort`
```
alsuren, line 22: When I was looking at this, I don't think that this `uniq` ever did anything, because it is handed a single line with all of the crates on. `uniq` needs to be handed a sorted list of lines, and it will only deduplicate lines that are immediately after each other.

In order to make it work when we have blessed crates at the top of the file, I installed `uq`. This doesn't require the list to be sorted, and can do the right thing for `cargo-make`.

sagiegurari: thanks

alsuren:
Thanks again for working on this with me. It looks like Linux and Windows have managed to get down to `cargo-chef`, but osx was in the middle of building something else when we dumped 100 packages into the build queue, so it's out of sync https://github.com/alsuren/cargo-quickinstall/actions

Slightly poetically, cargo-make used to be in the middle of the build queue, and then we promoted it to the top, so it will probably be built last, (once the pipeline has gone through all of the rest of the packages and restarted itself). Shout on https://gitter.im/cargo-quickinstall/community if you want me to step in and kick off a build sooner.

sagiegurari:
na, its ok. thanks a lot for accepting the PRs. its not urgent to have cargo-make built. more possible solution for an issue someone raised.

alsuren:
Hrm. cargo-make just got built for macos and it's not producing valid binaries. I will look into this tomorrow.

PR #68, opened 2022-02-07
The --dry-run flag

### LINKED ISSUE #64: add --dry-run option, which just prints the curl | tar command that would be run and then exits successfully

progga:
The ```--dry-run``` flag prints just the "curl | tar" command run during the happy path scenario where a pre-built crate binary is already available.

Closes #64

### DIFF
--- a/cargo-quickinstall/src/args.rs
+++ b/cargo-quickinstall/src/args.rs
@@ -10,6 +10,7 @@ OPTIONS:
         --version <VERSION>         Specify a version to install
         --target <TRIPLE>           Install package for the target triple
         --no-fallback               Don't fall back to `cargo install`
+        --dry-run                   Print the \"curl | tar\" command that would be run to fetch the binary
     -V, --print-version             Print version info and exit
     -h, --help                      Prints help information
 ";
@@ -21,6 +22,7 @@ pub struct CliOptions {
     pub fallback: bool,
     pub print_version: bool,
     pub help: bool,
+    pub dry_run: bool,
 }
 
 pub fn options_from_env() -> Result<CliOptions, Box<dyn std::error::Error + Send + Sync + 'static>>
@@ -32,6 +34,7 @@ pub fn options_from_env() -> Result<CliOptions, Box<dyn std::error::Error + Send
         fallback: !args.contains("--no-fallback"),
         print_version: args.contains(["-V", "--print-version"]),
         help: args.contains(["-h", "--help"]),
+        dry_run: args.contains("--dry-run"),
         // WARNING: We MUST parse all --options before parsing positional arguments,
         // because .subcommand() errors out if handed an arg with - at the start.
         crate_name: crate_name_from_positional_args(args)?,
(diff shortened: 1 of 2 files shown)


cargo-quickinstall/src/args.rs
```diff
@@ -10,6 +10,7 @@ OPTIONS:
         --version <VERSION>         Specify a version to install
         --target <TRIPLE>           Install package for the target triple
         --no-fallback               Don't fall back to `cargo install`
+        --dry-run                   Print the \"curl | tar\" command that would be run to fetch the binary
     -V, --print-version             Print version info and exit
     -h, --help                      Prints help information
 ";
```
alsuren, line 16: if you turn this string constant into a "raw" string, like
```
pub const HELP: &str = r#"USAGE:
...
"#;
```
then you don't need to escape the quotes. Alternatively, use backticks, like the line above does.

cargo-quickinstall/src/main.rs
```diff
@@ -339,6 +339,11 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
         None => get_target_triple()?,
     };
 
+    if options.dry_run {
+        println!("[Dry run] curl --user-agent \"cargo-quickinstall client (alsuren@gmail.com)\" --location --silent --show-error --fail https://github.com/alsuren/cargo-quickinstall/releases/download/{crate_name}-{version}-{target}/{crate_name}-{version}-{target}.tar.gz | tar -xzvvf - -C ~/.cargo/bin", crate_name=crate_name, version=version, target=target);
```
alsuren, line 343: It's all ended up on one line, so I will write things out as tickboxes. In general, try to avoid writing long lines like this. You can use backslash to escape the newline and subsequent indent if you need to (I can't find the docs for it, but https://stackoverflow.com/a/29483453)

- [ ] could we run the curl command (or with the --head argument to curl added as well) to check that the package is available before printing this?
- [ ] it is possible to make Command print itself in a way that is safe to paste into bash (using its Debug implementation). Could we refactor `curl_bytes(url)` to give us a `curl_command(url)` helper function. This would allow us to share more code (see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ec2ba6d47b709a065140c8bee4b773b4 for an example)? The tar command might be more difficult to share code for though (see below).
- [ ] (optional) Is `~/.cargo/bin` guaranteed to be where the user's cargo bin lives (even on windows)? That said, my main target audience for this feature is people who want to paste this into their CI pipelines (see the install instructions for https://github.com/cargo-quick/kache as an example). I'm expecting these users to be able to work out where they need to untar to, and this has the potential to add quite a lot of complexity, so I'm happy to leave this as it is.

I realise that I didn't write up this ticket very well. Sorry about that. Feel free to push back against all of those points and I'll happily merge this as-is.

alsuren (commented):
Thanks again for joining me on the hack and learn, and thanks for implementing this feature. 

I've left some suggestions, but all of them are optional (and you might get halfway through implementing one of them and decide that it's a bad idea). Please tell me which ones are good ideas and which are bad/tedious ideas, and I will happily merge without the bad ideas.

cargo-quickinstall/src/args.rs
Adnan-cds: I am going with the easy option, the backtick :)

cargo-quickinstall/src/main.rs
progga: I have implemented the above.  Please take a look.

alsuren: It's a thing of beauty

PR #75, opened 2022-02-28
Integration test proposal

progga:
Hi David,
As you have explained already, there is a smoke test, so I don't need to do anything about that.  Instead I have tried to cover only the *dry run* process in the integration test.  This has meant moving everything to a library crate.  Not sure if that's too much of a change.  So feel free to ignore this PR.

I have assumed that Ripgrep **v13.0.0**'s Linux x86_64 build hosted by cargo-quickinstall is permanent.  Do let me know if that's not the case.


cargo-quickinstall/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+    );
+    std::thread::spawn(move || {
+        // warehouse-clerk is known to return 404. This is fine. We only use it for stats gathering.
+        curl_head(&stats_url).unwrap_or_default();
+    })
+}
+
+pub fn do_dry_run(crate_details: &CrateDetails) -> String {
+    let result: String;
+
```
alsuren, line 115:
```suggestion
```

cargo-quickinstall/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+        let cargo_bin_dir = home::cargo_home().unwrap().join("bin");
+        let cargo_bin_dir_str = cargo_bin_dir.to_str().unwrap();
+        result = format!(
+            "{curl_cmd:?} | {untar_cmd:?}",
+            curl_cmd = prepare_curl_bytes_cmd(&crate_download_url),
+            untar_cmd = prepare_untar_cmd(cargo_bin_dir_str)
+        );
+    } else {
+        let cargo_install_cmd = prepare_cargo_install_cmd(crate_details);
+        result = format!("{:?}", cargo_install_cmd);
```
alsuren, line 133:
```suggestion
        format!("{:?}", cargo_install_cmd)
```

cargo-quickinstall/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    result
```
alsuren, line 136:
```suggestion
```

cargo-quickinstall/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+        target = crate_details.target
+    );
+    if curl_head(&crate_download_url).is_ok() {
+        let cargo_bin_dir = home::cargo_home().unwrap().join("bin");
+        let cargo_bin_dir_str = cargo_bin_dir.to_str().unwrap();
+        result = format!(
+            "{curl_cmd:?} | {untar_cmd:?}",
+            curl_cmd = prepare_curl_bytes_cmd(&crate_download_url),
+            untar_cmd = prepare_untar_cmd(cargo_bin_dir_str)
+        );
```
alsuren, line 130:
```suggestion
        format!(
            "{curl_cmd:?} | {untar_cmd:?}",
            curl_cmd = prepare_curl_bytes_cmd(&crate_download_url),
            untar_cmd = prepare_untar_cmd(cargo_bin_dir_str)
        )
```

cargo-quickinstall/src/lib.rs
alsuren: `if` statements are expressions in rust, so you can return directly from them. There's no need for the `result` variable.

cargo-quickinstall/tests/integration_tests.rs
```diff
@@ -0,0 +1,34 @@
...
+/// exist.
+#[test]
+fn do_dry_run_for_ripgrep() {
+    let crate_details = CrateDetails {
+        crate_name: "ripgrep".to_string(),
+        version: "13.0.0".to_string(),
+        target: "x86_64-unknown-linux-gnu".to_string(),
+    };
+
+    std::env::set_var("CARGO_HOME", "/tmp/mock-cargo-root");
```
alsuren, line 29: It's fine for this PR, but be aware that `std::env::set_var()` is not thread safe. If another test sets it to something else, you might get unexpected results.

If it becomes a problem later, I think the standard solution is to create a shared init() function that is called at the top of each test, and sets up the env the same way for all tests (potentially protected behind a std::sync::Once::call_once(||{}))

alsuren (approved):
I think this looks good. You're slowly polishing my terrible hacked-together-as-a-PoC code into something really solid.

I ended up reviewing this code with `git  blame -C cargo-quickinstall/src/lib.rs | less +/2022-02-28` because there were functional edits intermingled with the big refactor to move code around. Try to keep these things separate when you can.

Thanks again for your contributions.

cargo-quickinstall/src/lib.rs
progga: Done.

cargo-quickinstall/src/lib.rs
progga: Done.

cargo-quickinstall/tests/integration_tests.rs
progga: > but be aware that std::env::set_var() is not thread safe.

I have added another test covering ```install_crate()``` and this thread safety immediately became an issue.  I managed to avoid it for now by dropping the second call to ```std::env::set_var()```, but yes, we need a better fix.  The ```std::sync::Once```-based approach may not work as [mktemp](https://docs.rs/mktemp/latest/mktemp/) removes its tmp directory when it goes out of scope.  Perhaps we will just have to acquire a lock over the CARGO_HOME env variable.  Something for later.

On the upside, the test code looks Windows-proof although I haven't tested it there :)

alsuren (approved):

alsuren:
Thanks again for your contributions. Sorry I've been a bit slow reviewing this one. I will probably be a bit slow reviewing PRs this month in general, because I am a bit low on energy at the moment (and all of the spare energy I have for OSS is getting poured into https://github.com/cargo-quick/cargo-quick/projects/1?fullscreen=true ).

progga:
No problem David :)  These things can wait.  I hope you reach your normal energy level soon.  Cargo-quick looks useful.  I will have a look.

PR #78, opened 2022-03-12
Add license files for MIT and Apache 2.0

### LINKED ISSUE #76: Add LICENSE files to repo
I see `cargo-quickinstall`'s `Cargo.toml` lists `"MIT OR Apache-2.0"` for the license. It would be nice to have the LICENSE files added to the repo as well, just to make things more clear

LovecraftianHorror:
Closes #76 

This adds the license files listed in `cargo-quickinstall/Cargo.toml` to the repo along with a short notice indicating this at the bottom of the README.

I used 2020 as the starting copyright date since that's how far back the original commit adding the licenses to the `Cargo.toml` dates back (d2b05ea0442265e23f7aaab079a46d680a97229f). I also used the generic `cargo-quickinstall developers` since that seems to be a common convention instead of individually listing every contributor

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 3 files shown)


alsuren (approved):
Thanks for doing this. I should probably do the same for my other repos at some point. It's mostly my housemate that does the license dance, because his employer requires it.

Issue #49, opened 2022-01-22
[red team] Attempt to escape github actions executor and upload an artefact


alsuren:

alsuren:
I *think* that this is now solved. I will leave this "in review" until I can actually verify that it leaves me with a properly neutered token, and also that it's not possible to upload artifacts to workflow runs that are not the one your job is for.

Things to try would be:

- [ ] spin up a GitHub runner (https://github.com/actions/runner) inside a VM on a laptop, or in the cloud somewhere.
- [ ] Add a bunch of logging
- [ ] make a new repo with a job that just prints the token and sleeps forever, and assign it to that runner
- [ ] take the token and try to do stuff with it
- [ ] trigger another workflow run in parallel on a different runner (that also sleeps forever)
- [ ] try to use actions/upload-artifact to upload from one workflow run to the other

alsuren:
As part of this ticket, I think that I would also like to go around and trim down the permissions used in all other jobs. The cronjob and upload job both currently use a deploy key that is a bit overpowered. It would be nice to have at least the upload job go back to using a less powerful key.

PR #86, opened 2022-07-24
revoke permissions for builder

alsuren:
I think that this solves https://github.com/alsuren/cargo-quickinstall/issues/49 without any added complexity. 

I stumbled across this setting when trying to set up the webhook based scheme described in #49.
The paper trail went something like: https://github.com/organizations/cargo-quick/settings/actions
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

This allowed me to know that it was possible at the top level, and then I jumped into vscode to see whether it was possible to use it at the job level (using the red hat yaml plugin, which pointed me at https://json.schemastore.org/github-workflow.json)

Turns out it is: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions

### DIFF
--- a/.github/workflows/build-package.yml.template
+++ b/.github/workflows/build-package.yml.template
@@ -13,6 +13,7 @@ jobs:
   build-popular-package:
     name: Build
     runs-on: $BUILD_OS
+    permissions: {}
     defaults:
       run:
         shell: bash


alsuren:
Don't worry about the failing test. This only exists because I have a special case for the `actions` branch, which happens to have a race in it.

PR #88, opened 2022-08-19
tar defaults to format=gnu, whose sparse extension breaks the tar crate

alsuren:
See https://github.com/alsuren/cargo-quickinstall/issues/87 for discussion.

### DIFF
--- a/build-version.sh
+++ b/build-version.sh
@@ -57,6 +57,6 @@ done
 #
 # TODO: maybe we want to make a ~/.cargo-quickinstall/bin to untar into,
 # and add symlinks into ~/.cargo/bin, to aid debugging?
-tar -czf "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" $BINARIES
+tar --posix -czf "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" $BINARIES
 
 echo "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"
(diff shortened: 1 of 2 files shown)


build-version.sh
```diff
@@ -57,6 +57,6 @@ done
 #
 # TODO: maybe we want to make a ~/.cargo-quickinstall/bin to untar into,
 # and add symlinks into ~/.cargo/bin, to aid debugging?
-tar -czf "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" $BINARIES
+tar --posix -czf "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" $BINARIES
```
NobodyXu, line 60: We might have to use `--portability` here:

```suggestion
tar --portability -czf "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" $BINARIES
```

NobodyXu (changes requested):

build-version.sh
alsuren: Amusingly, I couldn't see --portability in the bsdtar man page.

NobodyXu: Perhaps we can use `--format=v7` instead (`--portability` seems to be same as it)?

NobodyXu (approved):

alsuren:
ugh. I cheated myself out of having code coverage for this change: https://github.com/alsuren/cargo-quickinstall/runs/7915295137?check_suite_focus=true#step:4:49

PR #93, opened 2022-09-16
Enable dependabot

NobodyXu:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# Dependabot dependency version checks / updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    # Workflow files stored in the
+    # default location of `.github/workflows`
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo"
+    directory: "/cargo-quickinstall"
+    schedule:
+      interval: "daily"


alsuren:
The Rust side of this only has 3 dependencies or something, so hopefully this won't be too noisy ;-).

The GitHub action side of it won't pick up anything that is in the template. Can you think of an easy way to make that work, or should we just deal with that later?

Also, do you want me to give you merge access, in case I fall behind with dependabot reviews?

alsuren (approved):

NobodyXu:
> The GitHub action side of it won't pick up anything that is in the template. Can you think of an easy way to make that work, or should we just deal with that later?

Hmmm, maybe we can rename that template if it doesn't work?
I'm not sure, let's deal with this later.

> Also, do you want me to give you merge access, in case I fall behind with dependabot reviews?

I'm willing to help you on merging the dependabot PRs, but I'm not very familiar with this codebase though.

alsuren:
> I'm willing to help you on merging the dependabot PRs, but I'm not very familiar with this codebase though.

That's fine. As long as you care about the project (which you seem to) and you're not planning to do anything intentionally malicious, it'll work out okay. I've sent you an invite.

Eventually I'd like to get to a place where binstall people can jump in and make changes to quickinstall. This doesn't have to happen immediately though, and it doesn't have to be you if you don't want to. I just picked on you because you seem like you care, and I think that's more important than anything else.

NobodyXu:
@alsuren Thank you for trusting me!

PR #98, opened 2022-09-16
Enable dependabot for `/Cargo.lock`

NobodyXu:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,12 @@ updates:
     schedule:
       interval: "daily"
   - package-ecosystem: "cargo"
+    # This will update the Cargo.lock
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo"
+    # This will update the cargo-quickinstall/Cargo.toml
     directory: "/cargo-quickinstall"
     schedule:
       interval: "daily"


.github/dependabot.yml
```diff
@@ -9,6 +9,12 @@ updates:
     schedule:
       interval: "daily"
   - package-ecosystem: "cargo"
+    # This will update the Cargo.lock
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "cargo"
+    # This will update the cargo-quickinstall/Cargo.toml
```
NobodyXu, line 17: Enable dependabot for `cargo-quickinstall` does not update the `Cargo.lock` file, so I enabled dependabot for `/` here to also update `Cargo.lock` when updating to new major versions of dependencies.

With that said, enabling dependabot for lockfile will increase the frequency of dependabot PRs for every new release, though that also means `cargo-quickinstall` will be able to use more up-to-date dep.

alsuren: I chose to make it a workspace at the start, because I was thinking that I would also write the stats server and build scripts in Rust, but in practice I wrote build scripts in bash and stats server in typescript (in https://github.com/alsuren/warehouse-clerk-tmp/blob/master/pages/api/crate/%5Btarball%5D.ts). We could probably kill off the workspace and bring the crate up to the top level if we wanted.

In practice, I don't think we have any transitive dependencies at the moment, so the number of PRs should remain small. Let's see how it goes.

alsuren (approved):

PR #105, opened 2022-12-23
Force static linking for openssl

### LINKED ISSUE #84: Problems with dynamically linked binaries (e.g. no OpenSSL 3 compatibilty)
Some of the newer OSes (notably Ubuntu 22.04 LTS Jammy Jellyfish) only ship with OpenSSL 3.0 and do not have an OpenSSL 1.1 package. This results in an error when trying to run the executable built by this project as Ubuntu 20.04 which the executables here are built with ships with OpenSSL 1.1:
```
/home/ubuntu/.cargo/bin/cargo-install-update: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

This is problematic because different OSes will have different versions of libraries available. I feel like linking more things statically would be helpful here. To allow to support many currently used operating systems, it probably would make sense to limit the number of libraries that you link against dynamically.

For starters, statically linking OpenSSL would help with a bunch of packages already. This isn't necessarily that easy since you would need a static build of OpenSSL but I think it might be achievable by either building a custom Docker image with a static OpenSSL build (and other static libs in the future) or using some existing work such as:
- https://github.com/emk/rust-musl-builder
- https://github.com/clux/muslrust
- https://github.com/golddranks/rust_musl_docker

All of these are musl-based as statically linking with glibc is usually advised against because of NSS issues and licensing.

I'd like to also mention Python's manylinux images as the Python packaging community seems to have solved this problem of supporting mainstream distros with a single build: https://peps.python.org/pep-0600/#core-definition
They approached this problem a bit differently and the specification allows you to dynamically link against glibc as well as any dynamic libraries that are available in ALL mainstream distros with that (or newer) glibc version. Anything else still would need to be statically linked. Considering that you can't really give separate builds for people with different glibc versions like Python does, you would presumably have to choose a single minimum glibc version for which you want to build. Maybe it's not that great of an idea for solving this issue but I wanted to at least mention how the Python packaging community approached it. The upside of this would be that you would still be able to build against glibc and just would have to limit yourself to only dynamically linking glibc and the few libraries that are available in ALL mainstream distros with that minimum glibc version.

NobodyXu:
Fixed #84 

### DIFF
--- a/build-version.sh
+++ b/build-version.sh
@@ -26,14 +26,14 @@ then
     # Compiling against musl libc is failing despite installing the musl-tools
     # deb.  Falling back to Rust's Alpine container whose default target
     # is x86_64-unknown-linux-musl.
-    podman run --name=official-alpine-rust docker.io/library/rust:alpine sh -c "apk add build-base; CARGO_PROFILE_RELEASE_CODEGEN_UNITS=\"1\" CARGO_PROFILE_RELEASE_LTO=\"fat\" cargo install $CRATE --version $VERSION"
+    podman run --name=official-alpine-rust docker.io/library/rust:alpine sh -c "apk add build-base; CARGO_PROFILE_RELEASE_CODEGEN_UNITS=\"1\" CARGO_PROFILE_RELEASE_LTO=\"fat\" OPENSSL_STATIC=1 cargo install $CRATE --version $VERSION"
     podman cp official-alpine-rust:/usr/local/cargo "${TEMPDIR}/"
 
     CARGO_BIN_DIR="${TEMPDIR}/cargo/bin"
     CRATES2_JSON_PATH="${TEMPDIR}/cargo/.crates2.json"
 else
     rustup target add "$TARGET_ARCH"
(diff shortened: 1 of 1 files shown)


alsuren (approved):
Sure. Let's give this a go.

I have previously been reluctant to rebuild packages (I like the illusion of immutability). If you want to force a rebuild, you can:
* Pick a package
* Delete the tag (this will also delete the release assets)
* make sure the package has recently been requested (stats reset at midnight UTC) or is in popular-packages.txt
* Wait a few hours, or (from memory: in typing on my phone) run trigger-package-build.sh a couple of times and keep your eye on the github actions tab (this script is really slow because it rate limits itself when calling out to the crates.io API. You can force it to only check a single architecture if you set an environment variable. You can also influence which package it starts checking at)

NobodyXu:
@alsuren Thanks for the suggestion, I also don't like deleting existing builds, though I'd love to test this patch on in wild and see whether there's any bug in it.

Issue #26, opened 2022-01-22
[tiny] document how to read the github actions build logs


alsuren:
At the moment, https://github.com/cargo-bins/cargo-quickinstall/actions is quite messy because we don't implement "cancelled" builds properly. We should fix this first.

The search also basically doesn't work, because it is based on the filename of the yaml file. We use a different filename for each arch, but don't include the name of the package in this.

- [x] make it so that pushing to the `actions` branch isn't racey (currently one of them succeeds and the other fails) - Maybe only trigger one of them and not the other?
  - [ ] consider moving the cronjob and weekly-recheck triggers from the `actions` branch to a PR trigger that triggers whenever you make a PR against files in the .github ? (this might be tricky, because we currently substitute in the name of the branch when instantiating the template, and that might be ephemeral on PR builds)
- [x] implement cancelling of builds properly, so they show up as "cancelled" in the logs

PR #107, opened 2022-12-27
misc build Actions cleanups

alsuren:
I thought I might as well clean some long-serving uglinesses up before I hand them over. Hopefully the code review process will act as a way for @NobodyXu to learn how things work ;-)

This vaguely relates to #26 I guess.

Probably best to review things commit-by-commit. I didn't write them in that order, but the commits should be self-contained after my rebasing.

pushing to the actions branch (with a PR open against it) now triggers the following cascade (oldest at the bottom):

Screenshot 2022-12-27 at 23 49 02

(you can ignore the `Automated Tests` and `Self-Build` actions)

* "weekly recheck" is triggered (as a hack) 
* it runs `make recheck` which runs `trigger-package-build.sh`
* `trigger-package-build.sh` fills in .github/workflows/build-package.yml.template a bunch of times, and pushes each to the appropriate trigger/$arch branch. It sets FORCE=1 so these pushes are likely to be empty commits (`git commit --allow-empty`)
* Each push triggers a "Build for $arch" build for cargo-quickinstall, because there is nothing else to build (it only checks 10 packages anyway, because waiting 20 minutes for every change on the `actions` branch was getting really boring)
* each build quickly realises that cargo-quickinstall is already packaged, so there is nothing to do. It then cancels itself (previously it was set up to cancel itself, but it would die before getting to that step).

### DIFF
--- a/.github/workflows/cronjob.yml
+++ b/.github/workflows/cronjob.yml
@@ -1,13 +1,9 @@
 name: Cronjob
 
 # Run every hour on the hour.
-# `actions` is the name of my prototyping branch, so also run this job when I push there.
 on:
   schedule:
     - cron: "0 * * * *"
-  push:
-    branches:
-      - actions
 
 jobs:
   build-popular-package:
--- a/.github/workflows/selfbuild.yml
+++ b/.github/workflows/selfbuild.yml
@@ -35,14 +35,14 @@ jobs:
         run: |
           set -euxo pipefail
           touch .env
-          VERSION=$(
+          VERSION="$(
             curl \
               --user-agent "cargo-quickinstall build pipeline (alsuren@gmail.com)" \
               --fail "https://crates.io/api/v1/crates/cargo-quickinstall" \
               | jq -r '.versions[0].num'
-          )
-          env VERSION=$VERSION TARGET_ARCH=$TARGET_ARCH ./build-version.sh cargo-quickinstall
-          env ALWAYS_BUILD=1 VERSION=$VERSION TARGET_ARCH=$TARGET_ARCH ./build-version.sh cargo-quickinstall
+          )"
+          VERSION="$VERSION" TARGET_ARCH="$TARGET_ARCH" ./build-version.sh cargo-quickinstall
+          ALWAYS_BUILD=1 VERSION="$VERSION" TARGET_ARCH="$TARGET_ARCH" ./build-version.sh cargo-quickinstall
       - name: Install Thyself
         run: cargo install --path cargo-quickinstall
       - name: Install Thyself with Thyself (or fallback to sensei on windows)
--- a/.github/workflows/selfbuild.yml
+++ b/.github/workflows/selfbuild.yml
@@ -39,7 +39,7 @@ jobs:
             curl \
               --user-agent "cargo-quickinstall build pipeline (alsuren@gmail.com)" \
               --fail "https://crates.io/api/v1/crates/cargo-quickinstall" \
-              | jq -r .versions[0].num
+              | jq -r '.versions[0].num'
           )
           env VERSION=$VERSION TARGET_ARCH=$TARGET_ARCH ./build-version.sh cargo-quickinstall
           env ALWAYS_BUILD=1 VERSION=$VERSION TARGET_ARCH=$TARGET_ARCH ./build-version.sh cargo-quickinstall
--- a/build-version.sh
+++ b/build-version.sh
@@ -60,6 +60,6 @@ done
 #
 # BINARIES is a space-separated list of files, so it can't be quoted
 # shellcheck disable=SC2086
-tar --format=v7 -czf "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" $BINARIES
+tar --format=v7 -c $BINARIES | gzip -9 -c >"${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"
 
 echo "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"
--- a/build-version.sh
+++ b/build-version.sh
@@ -40,14 +40,14 @@ else
 fi
 
 BINARIES=$(
-    cat $CRATES2_JSON_PATH | jq -r '
-        .installs | to_entries[] | select(.key|startswith("'${CRATE}' ")) | .value.bins | .[]
-    ' | tr '\r' ' '
+    jq -r '
+        .installs | to_entries[] | select(.key|startswith("'"${CRATE}"' ")) | .value.bins | .[]
+    ' "$CRATES2_JSON_PATH" | tr '\r' ' '
 )
 
-cd $CARGO_BIN_DIR
+cd "$CARGO_BIN_DIR"
 for file in $BINARIES; do
-    if file $file | grep ': data$'; then
+    if file "$file" | grep ': data$'; then
         echo "something wrong with $file. Should be recognised as executable."
         exit 1
     fi
@@ -57,6 +57,9 @@ done
 #
 # TODO: maybe we want to make a ~/.cargo-quickinstall/bin to untar into,
 # and add symlinks into ~/.cargo/bin, to aid debugging?
+#
+# BINARIES is a space-separated list of files, so it can't be quoted
+# shellcheck disable=SC2086
 tar --format=v7 -czf "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" $BINARIES
 
 echo "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"
--- a/cargo-quickinstall/src/lib.rs
+++ b/cargo-quickinstall/src/lib.rs
@@ -178,6 +178,7 @@ fn curl_head(url: &str) -> Result<Vec<u8>, InstallError> {
         .arg("--silent")
         .arg("--show-error")
         .arg("--fail")
+        .arg("--location")
         .arg(url)
         .output()?;
     if !output.status.success() {
--- a/next-unbuilt-package.sh
+++ b/next-unbuilt-package.sh
@@ -30,7 +30,8 @@ POPULAR_CRATES=$(
       grep -v '/' |
       grep -A1000 --line-regexp "${START_AFTER_CRATE:-.*}" |
       # drop the first line (the one that matched)
-      tail -n +2 ||
+      tail -n +2 |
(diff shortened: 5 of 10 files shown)


.github/workflows/cronjob.yml
```diff
@@ -1,13 +1,9 @@
 name: Cronjob
 
 # Run every hour on the hour.
-# `actions` is the name of my prototyping branch, so also run this job when I push there.
 on:
   schedule:
     - cron: "0 * * * *"
-  push:
-    branches:
-      - actions
```
alsuren, line 10: The actions branch is still magical, but now it only triggers the `make recheck` workflow (in another file)

cargo-quickinstall/src/lib.rs
```diff
@@ -178,6 +178,7 @@ fn curl_head(url: &str) -> Result<Vec<u8>, InstallError> {
         .arg("--silent")
         .arg("--show-error")
         .arg("--fail")
+        .arg("--location")
```
alsuren, line 181: This is unrelated, but I had a failing test on my PR and this is the fix (another fix would be to point it at the right repo so there aren't any redirects in the failure case - we should definitely do that too, and make a release. --dry-run is very niche though, so let's not worry about it being broken for now).

next-unbuilt-package.sh
```diff
@@ -48,10 +49,14 @@ for CRATE in $POPULAR_CRATES; do
...
-  curl_slowly --location --fail "https://crates.io/api/v1/crates/${CRATE}" >"$TEMPDIR/crates.io-response.json"
-  VERSION=$(cat "$TEMPDIR/crates.io-response.json" | jq -r .versions[0].num)
-  LICENSE=$(cat "$TEMPDIR/crates.io-response.json" | jq -r .versions[0].license | sed -e 's:/:", ":g' -e 's/ OR /", "/g')
+  RESPONSE_DIR="$TEMPDIR/crates.io-responses/"
+  RESPONSE_FILENAME="$RESPONSE_DIR/$CRATE.json"
+  if [[ ! -f "$RESPONSE_FILENAME" ]]; then
+    mkdir -p "$RESPONSE_DIR"
+    curl_slowly --location --fail "https://crates.io/api/v1/crates/${CRATE}" >"$RESPONSE_FILENAME"
+  fi
+  VERSION=$(cat "$RESPONSE_FILENAME" | jq -r .versions[0].num)
```
NobodyXu, line 58:
```suggestion
  VERSION=$(jq -r '.versions[0].num' "$RESPONSE_FILENAME")
```

NobodyXu, line 58:
```suggestion
  VERSION=$(jq -r '.versions[0].num' "$RESPONSE_FILENAME")
```

NobodyXu, line 58:
```suggestion
  VERSION=$(jq -r '.versions[0].num' "$RESPONSE_FILENAME")
```

next-unbuilt-package.sh
```diff
@@ -48,10 +49,14 @@ for CRATE in $POPULAR_CRATES; do
...
+  LICENSE=$(cat "$RESPONSE_FILENAME" | jq -r .versions[0].license | sed -e 's:/:", ":g' -e 's/ OR /", "/g')
```
NobodyXu, line 59:
```suggestion
  LICENSE=$(jq -r '.versions[0].license' "$RESPONSE_FILENAME" | sed -e 's:/:", ":g' -e 's/ OR /", "/g')
```

NobodyXu, line 59:
```suggestion
  LICENSE=$(jq -r '.versions[0].license' "$RESPONSE_FILENAME" | sed -e 's:/:", ":g' -e 's/ OR /", "/g')
```

NobodyXu (changes requested):

next-unbuilt-package.sh
NobodyXu: Due to some internal github error, I created multiple duplicate reviews, closing this as duplicate.

NobodyXu: Due to some internal github error, I created multiple duplicate reviews, closing this as duplicate.

next-unbuilt-package.sh
NobodyXu: Due to some internal github error, I created multiple duplicate reviews, closing this as duplicate.

next-unbuilt-package.sh
alsuren: You will notice that I tend to use `cat | ...` a lot. I find that it reduces cognitive load, because tools are only used in their "filter" mode, and information flow is more uniformly left to right throughout the codebase. 

As an added bonus, `cat | jq` is faster for prototyping on the command line, because you are always editing the last word on the line each time you iterate.

Shellcheck also dislikes `cat | ...`, so I am probably alone in my preference. I will do a pass to make things more best-practice compliant after breakfast.

build-version.sh
```diff
@@ -40,14 +40,14 @@ else
...
+    jq -r '
+        .installs | to_entries[] | select(.key|startswith("'"${CRATE}"' ")) | .value.bins | .[]
+    ' "$CRATES2_JSON_PATH" | tr '\r' ' '
 )
 
-cd $CARGO_BIN_DIR
+cd "$CARGO_BIN_DIR"
 for file in $BINARIES; do
-    if file $file | grep ': data$'; then
+    if file "$file" | grep ': data$'; then
```
NobodyXu, line 50: Could use this instead:

```suggestion
    if [ -f "$file" ] && [ -x "$file" ]; then
```

The first `[ -f "$file" ]` checks to make sure `$file` is a regular file while `[ -x "$file" ]` checks to make sure its executable bit is set.

build-version.sh
```diff
@@ -57,6 +57,9 @@ done
 #
 # TODO: maybe we want to make a ~/.cargo-quickinstall/bin to untar into,
 # and add symlinks into ~/.cargo/bin, to aid debugging?
+#
+# BINARIES is a space-separated list of files, so it can't be quoted
+# shellcheck disable=SC2086
 tar --format=v7 -czf "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" $BINARIES
```
NobodyXu, line 63: I'm thinking perhaps we can force the highest compression rate here?

```suggestion
tar --format=v7 -c $BINARIES | gzip -9 -c > "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"
```

I'm not sure whether `tar` uses the highest compression level or not.

.github/workflows/selfbuild.yml
```diff
@@ -39,7 +39,7 @@ jobs:
             curl \
               --user-agent "cargo-quickinstall build pipeline (alsuren@gmail.com)" \
               --fail "https://crates.io/api/v1/crates/cargo-quickinstall" \
-              | jq -r .versions[0].num
+              | jq -r '.versions[0].num'
           )
           env VERSION=$VERSION TARGET_ARCH=$TARGET_ARCH ./build-version.sh cargo-quickinstall
```
NobodyXu, line 44:
```suggestion
         VERSION="$VERSION" TARGET_ARCH="$TARGET_ARCH" ./build-version.sh cargo-quickinstall
```

.github/workflows/selfbuild.yml
```diff
@@ -39,7 +39,7 @@ jobs:
...
           env ALWAYS_BUILD=1 VERSION=$VERSION TARGET_ARCH=$TARGET_ARCH ./build-version.sh cargo-quickinstall
```
NobodyXu, line 45:
```suggestion
         ALWAYS_BUILD=1 VERSION="$VERSION" TARGET_ARCH="$TARGET_ARCH" ./build-version.sh cargo-quickinstall
```

NobodyXu (changes requested):

build-version.sh
NobodyXu: I just realize that this script might need to run on windows, which doesn't have the executable bit, so probably better to go with the original one.

alsuren: the problem that this is trying to catch is that macos runners sometimes don't understand the binaries that had been built (my computer also didn't understand them). The executable bit wouldn't help us here.

I have no idea why this was happening, or whether it still happens, but asking `file` seemed to be enough to catch it.

build-version.sh
alsuren: I have taken the philosophy that you get multiple orders of magnitude by installing from precompiled binaries, so I haven't gone chasing after marginal compression gains.

I also wasn't sure whether the windows runners had gzip as a separate tool (I was surprised to find that `tar` is available by default on windows these days). If you think it is, I'm happy to go with this.

.github/workflows/selfbuild.yml
alsuren: heh. I can't remember what got me into the habit of writing `env` explicitly. It might be from writing portable npm scripts for running on windows or something.

alsuren:
can you tell that I hacked most of this together in a few days of downtime at work? ;-)

build-version.sh
NobodyXu: > I also wasn't sure whether the windows runners had gzip as a separate tool (I was surprised to find that tar is available by default on windows these days). If you think it is, I'm happy to go with this.

I'm checking, but AFAIK `tar` simply spawns a new command for the compression/decompression for gz and other command, so it should be available.

NobodyXu: I didn't find `gzip` or `tar` on https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md , perhaps we can simply test it out on CI?

NobodyXu:
> can you tell that I hacked most of this together in a few days of downtime at work? ;-)

You are amazing.

build-version.sh
alsuren: oooh! we should install one of those tmux debugger actions that we used at my previous job - https://github.com/marketplace/actions/debugging-with-tmate

NobodyXu: The CI on windows runs just fine, so I will close this.

NobodyXu: > oooh! we should install one of those tmux debugger actions that we used at my previous job - https://github.com/marketplace/actions/debugging-with-tmate

I have no idea such thing exists, I debugged my GHA script by pushing new commits...

alsuren: note that this won't actually get exercised until we try to build a package that's not cancelled. We will definitely be building a package for cargo-quickinstall when we do our release, so I'm fine to merge this to main and revert it then if it causes problems.

NobodyXu (approved):

build-version.sh
alsuren: ooh! I am cleverer than I give myself credit for: selfbuild *does* exercise this codepath on all PRs.

Issue #58, opened 2022-01-22
move this repo into the cargo-quick org


alsuren:
- [ ] check what happens to github releases artifacts by creating a test repo and moving it into the cargo-quick org
- [ ] update scripts to use the new github releases url
- [ ] make a release that uses the new github releases url but falls back to the old place
- [ ] move the repo
- [ ] remove the falback and make another release

PR #108, opened 2022-12-27
update to point at cargo-bins org

alsuren:
Since github automatically does http redirects to the new org, there's very little left to do on #58 .

Once this is merged, we should probably make a release https://github.com/cargo-bins/cargo-quickinstall#releasing but there's no urgency - everything seems to be trucking along as usual. The only thing that broke with the rename was that `--dry-run` thinks that we have a package for everything.

I can do the release + announcement if you want, to announce the handover.

I think the only thing that's left after this is rotating secrets.CRONJOB_DEPLOY_KEY

### DIFF
--- a/.github/workflows/build-package.yml.template
+++ b/.github/workflows/build-package.yml.template
@@ -2,7 +2,7 @@ name: Build for $TARGET_ARCH
 
 # Triggered by push from cronjob.
 # $CRATE, $VERSION, $TARGET_ARCH, $BRANCH and $BUILD_OS are replaced using sed.
-# Title of job in https://github.com/alsuren/cargo-quickinstall/actions
+# Title of job in https://github.com/cargo-bins/cargo-quickinstall/actions
 # is set by the commit message, which is created by trigger-package-build.sh
 on:
   push:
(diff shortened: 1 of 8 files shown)


NobodyXu (approved):

NobodyXu:
> The only thing that broke with the rename was that --dry-run thinks that we have a package for everything.

I tested `cargo-binstall` with `--strategies quick-install`, seems like somehow it's not affected by this.

> I can do the release + announcement if you want, to announce the handover.

That will be great!

BTW, would you like to stay in the cargo-bins to reduce the bus factor for binstall and quickinstall?

alsuren:
> BTW, would you like to stay in the cargo-bins to reduce the bus factor for binstall and quickinstall?

In the long term, I'd prefer to hand over the quickinstall stuff and only be subscribed to @mentions (I'm happy to remain in the org in case a real catastrophe happens). I realise that we're a long way off from that though - there is a lot of cleaning up to do before that. I'm quite enjoying the process of cleaning things up, so I will stick at it. I will also include the "maintainers wanted" call as part of my release announcement.

Issue #46, opened 2022-01-22
thinking about converting build scripts to python?


alsuren:
Another alternative would be google/zx

PR #109, opened 2022-12-28
use shfmt, and enforce in CI

alsuren:
I have split this out as a separate PR because it's super-noisy. It's almost entirely indentation changes though, because at some point my vscode decided that one of the files should be indented with 2 spaces and I didn't notice.

- [x] install shellcheck in ci if it's not already there
- [x] change base from actions to main once #107 is merged

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*.sh]
+indent_style = space
+indent_size = 4 
--- a/.github/workflows/selfbuild.yml
+++ b/.github/workflows/selfbuild.yml
@@ -2,8 +2,6 @@ name: Self-Build
 
 on:
   pull_request:
-    branches:
-      - main
 
 jobs:
   self-build:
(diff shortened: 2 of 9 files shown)


NobodyXu (approved):

NobodyXu:
There are so many bash scripts and the use of `jq` and `sed` is a bit hard to read for me, probably because I'm not good at `jq` and not familiar with advanced usage of `sed`.

Will it be a good idea to use python script or even rust for that?

alsuren:
> Will it be a good idea to use python script or even rust for that?

Potentially, yes. #46 discusses some options. I quite like the idea of porting them all to use zx, but python would be fine.

Rust would also be fun (and the repo is already set up as a workspace, so adding an xtask package would be easy enough). We probably wouldn't want to compile the xtask package from scratch every time. This means we would have to think carefully about our cacheing scheme if we want to avoid giving untrusted runners write access to a cache that others are reading from.

Which language would you pick?

NobodyXu:
> Which language would you pick?

I might as well go with python to avoid all the hassle of compilation.

While it's not as good as rust, it's still way better than bash IMHO.

PR #116, opened 2023-01-11
Use `cargo-binstall` instead of `curl | tar` by default

### LINKED ISSUE #71: cargo-quickinstall --uninstall packagename
* re-fetch the binary package
* run it through tar to list the contents
* delete each file in the tar archive from bin/.

NobodyXu:
Fixed #71

### DIFF
--- a/cargo-quickinstall/src/main.rs
+++ b/cargo-quickinstall/src/main.rs
@@ -109,7 +109,7 @@ fn do_main_binstall(
     version: Option<String>,
     target: Option<String>,
     dry_run: bool,
-    fallback: bool,
+    _fallback: bool,
 ) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
     let is_binstall_compatible = get_cargo_binstall_version()
         .map(
@@ -129,10 +129,12 @@ fn do_main_binstall(
         .unwrap_or(false);
 
     if !is_binstall_compatible {
-        do_main_curl("cargo-binstall".to_string(), None, None, dry_run, fallback)?;
+        do_main_curl("cargo-binstall".to_string(), None, None, dry_run, false)?;
 
         if !dry_run {
-            println!("Bootstrapping cargo-binstall with itself to make `cargo uninstall ` work properly");
+            println!(
+                "Bootstrapping cargo-binstall with itself to make `cargo uninstall ` work properly"
+            );
             do_install_binstall(
                 vec![Crate {
                     name: "cargo-binstall".to_string(),
--- a/cargo-quickinstall/src/main.rs
+++ b/cargo-quickinstall/src/main.rs
@@ -132,7 +132,7 @@ fn do_main_binstall(
         do_main_curl("cargo-binstall".to_string(), None, None, dry_run, fallback)?;
 
         if !dry_run {
-            println!("Bootstrap cargo-binstall itself to update $CARGO_HOME/.crates.toml");
+            println!("Bootstrapping cargo-binstall with itself to make `cargo uninstall ` work properly");
             do_install_binstall(
                 vec![Crate {
                     name: "cargo-binstall".to_string(),
--- a/cargo-quickinstall/src/main.rs
+++ b/cargo-quickinstall/src/main.rs
@@ -187,9 +187,9 @@ fn do_install_binstall(
     if !status.success() {
         Err(format!(
             "`{} {}` failed with {status}",
-            cmd.get_program().to_string_lossy().into_owned(),
+            cmd.get_program().to_string_lossy(),
             cmd.get_args()
-                .map(|arg| arg.to_string_lossy().into_owned())
+                .map(|arg| arg.to_string_lossy())
                 .collect::<Vec<_>>()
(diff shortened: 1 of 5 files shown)


cargo-quickinstall/src/main.rs
```diff
@@ -31,30 +31,170 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
...
+            version,
+        }],
+        target,
+        dry_run,
+        false,
+    )
+}
+
+fn do_install_binstall(
+    crates: Vec<Crate>,
```
NobodyXu, line 162: `do_install_binstall` takes `Vec<Crate>` here since I plan on submiting another PR to support batching mode.

cargo-quickinstall/src/main.rs
```diff
@@ -31,30 +31,170 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
...
+                _ => false,
+            },
+        )
+        .unwrap_or(false);
+
+    if !is_binstall_compatible {
+        do_main_curl("cargo-binstall".to_string(), None, None, dry_run, fallback)?;
+
+        if !dry_run {
+            println!("Bootstrap cargo-binstall itself to update $CARGO_HOME/.crates.toml");
```
alsuren, line 135:
```suggestion
            println!("Bootstrapping cargo-binstall with itself to make `cargo uninstall ` work properly");
```

I haven't tried this out locally yet. would the full list of println messages for the bootstrapping flow make sense to someone who has not used rust?

cargo-quickinstall/src/main.rs
```diff
@@ -31,30 +31,170 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
...
+        dry_run,
+        false,
+    )
+}
+
+fn do_install_binstall(
+    crates: Vec<Crate>,
+    target: Option<String>,
+    dry_run: bool,
+    force: bool,
```
alsuren, line 165: Optional, because this is probably my fault in the first place. I'm increasingly way of functions that take multiple boolean positional args, because I've managed to get the order wrong on a few occasions now.

Does it make sense to supply --dry-run and --force together? If not, it might be worth making an `enum Mode { Normal, Force, DryRun }` or something.

Otherwise, package them up into a struct might be worthwhile.

cargo-quickinstall/src/main.rs
```diff
@@ -31,30 +31,170 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
...
+                                .map(|minor| minor >= 17)
+                                .unwrap_or(false)
+                }
+                _ => false,
+            },
+        )
+        .unwrap_or(false);
+
+    if !is_binstall_compatible {
+        do_main_curl("cargo-binstall".to_string(), None, None, dry_run, fallback)?;
```
alsuren, line 132: We should probably never install binstall from source because it might take longer than the thing they wanted in the first place. If we don't have it prebuilt then just print a warning and install the package the user asked for the old fashioned way.

cargo-quickinstall/src/main.rs
```diff
@@ -31,30 +31,170 @@ fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
...
+
+fn do_install_binstall(
+    crates: Vec<Crate>,
+    target: Option<String>,
+    dry_run: bool,
+    force: bool,
+) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
+    let mut cmd = std::process::Command::new("cargo");
+
+    cmd.arg("binstall").arg("--no-confirm");
```
alsuren, line 169: This is a change of behaviour - it will now pick the upstream package if binstall can find it.

I think this is probably the right thing to do as long as you did it on purpose.

alsuren (approved):
This is pretty much what I had in mind. Nice work. Feel free to ignore any comments you want.

cargo-quickinstall/src/main.rs
NobodyXu: > I haven't tried this out locally yet. would the full list of println messages for the bootstrapping flow make sense to someone who has not used rust?

Not sure, but I stilll think it's better to at least print something so that they know quick-install is not stuck and is still making progress.

cargo-quickinstall/src/main.rs
NobodyXu: Yes, I think using upstream package is better, but maybe I should add more doc for this.

cargo-quickinstall/src/main.rs
NobodyXu: > We should probably never install binstall from source because it might take longer than the thing they wanted in the first place.

That's right, I didn't think of that.
I also didn't know `fallback` is for `cargo-install` as I didn't check for that.

I guess it makes sense to set it to `false` then, as compiling `binstall `takes at least 1m10s on my M1.

NobodyXu: Done

PR #118, opened 2023-01-12
Tweak `profile.release` for size

NobodyXu:
On my M1, this increase the compilation time from 1.15s => 3.15s, but reduce the size from 792K => 412K.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,3 +2,10 @@
 members = [
     "cargo-quickinstall",
 ]
+
+[profile.release]
+opt-level = "z"
+lto = true
+codegen-units = 1
+panic = "abort"
+strip = "symbols"


NobodyXu:
The CI build time goes from 1m => 5m, indeed a huge jump but not unacceptable.

alsuren:
I don't think that the trade off is with it in this case. The CI time is probably more representative of what it is will feel than the compiler time on your M1. If you think that this is a red herring, and out users won't feel this slowdown then maybe it's fine.

My philosophy on optimisation for quickinstall had always been "we get multiple orders of magnitude improvements in speed, space and bandwidth by fetching precompiled tarballs, so let's not over complicate things by chasing after marginal gains". This is why I just picked tar's default compression scheme, etc. In this case, you're not only sacrificing bootstrap time - you're also theoretically sacrificing debugability potentially? (Although quickinstall is so simple that I can't imagine anyone ever needing to run it under gdb).

All that said, I'm not going to try to block this one if you want to merge it. You are doing an excellent job and I'm happy for you to have final say on all PRs.

NobodyXu:
That's true, reducing from 700K to 400K usually does not matter much for the end-users.
And since most users use `cargo-quickinstall` for bootstraping by compiling it on their own computers, it's probably better to keep the compilation as fast as possible.

I will close this PR.

PR #120, opened 2023-01-12
Optimize `report_stats_in_background`

NobodyXu:
- Refactor: Extract new fn `prepare_curl_head_cmd`
 - Optimize `report_stats_in_background`: Avoid `thread::spawn`
 - Optimize `report_stats_in_background`: Rm one `String` creation

### DIFF
--- a/cargo-quickinstall/src/lib.rs
+++ b/cargo-quickinstall/src/lib.rs
@@ -58,6 +58,7 @@ pub fn install_crate_curl(details: &CrateDetails, fallback: bool) -> Result<(), 
                 details.crate_name, details.version, details.target
             );
             println!("We have reported your installation request, so it should be built soon.");
+
             println!("Falling back to `cargo install`.");
 
             let status = prepare_cargo_install_cmd(details).status()?;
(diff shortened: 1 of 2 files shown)


alsuren (approved):
I assume that windows does the right thing here (I have no idea whether Windows does the right thing under the old scheme. They're probably equivalent)

Looks good. I never thought of doing that

cargo-quickinstall/src/lib.rs
```diff
@@ -182,17 +180,23 @@ fn untar(tarball: Vec<u8>) -> Result<String, InstallError> {
...
-        .arg("--user-agent")
+fn prepare_curl_head_cmd(url: &str) -> std::process::Command {
+    let mut cmd = std::process::Command::new("curl");
+
+    cmd.arg("--user-agent")
         .arg("cargo-quickinstall client (alsuren@gmail.com)")
         .arg("--head")
         .arg("--silent")
         .arg("--show-error")
         .arg("--fail")
```
alsuren, line 191: Out of interest: what happens if the stats call fails when it has been spawned (you could test this by turning off WiFi or hacking your /etc/hosts to point the stats ending somewhere bogus). Does init log anything about it anywhere or is it just a tree falling in the woods with nobody listening?

Also, I just sent you an email about the stats server because this PR reminded me.

NobodyXu: > Out of interest: what happens if the stats call fails when it has been spawned (you could test this by turning off WiFi or hacking your /etc/hosts to point the stats ending somewhere bogus). Does init log anything about it anywhere or is it just a tree falling in the woods with nobody listening?

I believe the init process would just `wait` for it so that it can clean up the zombies, it doesn't and can't log anything as it does not have access to its stdout or stderr.

> Also, I just sent you an email about the stats server because this PR reminded me.

Thanks, I received the email.
I would reply after dinner.

PR #121, opened 2023-01-13
Refactor, minor fix on panic during error and optimization

NobodyXu:


cargo-quickinstall/src/lib.rs
```diff
@@ -129,141 +143,112 @@ pub fn do_dry_run_curl(crate_details: &CrateDetails) -> String {
...
-        return Err(CommandFailed {
-            command: "tar -xzvvf - -C ~/.cargo/bin".to_string(),
-            stdout,
-            stderr,
-        }
-        .into());
-    }
+    // Propagate this error before Propagating error tar since
+    // if tar fails, it's likely due to curl failed.
+    curl.wait_with_output_checked_status()?;
```
alsuren, line 168: it's hard to see how you're going to handle the `err.is_curl_404()` case under this scheme.

cargo-quickinstall/src/utils.rs
```diff
@@ -0,0 +1,14 @@
...
+pub fn utf8_to_string_lossy(bytes: Vec<u8>) -> String {
+    String::from_utf8(bytes)
+        .unwrap_or_else(|err| String::from_utf8_lossy(err.as_bytes()).into_owned())
+}
+
+pub fn get_cargo_bin_dir() -> io::Result<PathBuf> {
+    cargo_home().map(|mut cargo_bin_dir| {
+        cargo_bin_dir.push("bin");
+        cargo_bin_dir
+    })
```
alsuren, line 13: (optional) I tend to find ? and ok-wrapping easier to read. The difference is more pronounced when things get more complicated (like with the get_owned() and_then() chain above), but consistency is also valuable.
```suggestion
    let mut cargo_bin_dir = cargo_home()?;
    cargo_bin_dir.push("bin");
    Ok(cargo_bin_dir)
```

also we don't care about performance here, so join is probably fine:

```suggestion
    let cargo_bin_dir = cargo_home()?;
    Ok(cargo_bin_dir.join("bin"))
```

cargo-quickinstall/tests/integration_tests.rs
```diff
@@ -36,9 +47,9 @@ fn do_dry_run_for_ripgrep() {
         target: "x86_64-unknown-linux-gnu".to_string(),
     };
 
-    let result = do_dry_run_curl(&crate_details);
+    let result = do_dry_run_curl(&crate_details).unwrap();
 
-    let expected_prefix = r#""curl" "--user-agent" "cargo-quickinstall client (alsuren@gmail.com)" "--location" "--silent" "--show-error" "--fail" "https://github.com/cargo-bins/cargo-quickinstall/releases/download/ripgrep-13.0.0-x86_64-unknown-linux-gnu/ripgrep-13.0.0-x86_64-unknown-linux-gnu.tar.gz" | "tar" "-xzvvf" "-" "-C""#;
+    let expected_prefix = "curl --user-agent cargo-quickinstall client (alsuren@gmail.com) --location --silent --show-error --fail https://github.com/cargo-bins/cargo-quickinstall/releases/download/ripgrep-13.0.0-x86_64-unknown-linux-gnu/ripgrep-13.0.0-x86_64-unknown-linux-gnu.tar.gz | tar -xzvvf - -C";
```
alsuren, line 52: as discussed above, this is a breaking change. You need to at least quote the user agent string because it contains spaces and brackets.

cargo-quickinstall/tests/integration_tests.rs
```diff
@@ -51,8 +62,16 @@ fn do_dry_run_for_nonexistent_package() {
...
-    let expected = r#""cargo" "install" "nonexisting_crate_12345" "--version" "99""#;
+    let expected = "cargo install nonexisting_crate_12345 --version 99";
     assert_eq!(expected, &result);
 }
+
+#[test]
+fn test_get_latest_version() {
+    assert_eq!(
+        get_latest_version("cargo-quickinstall").unwrap(),
+        env!("CARGO_PKG_VERSION")
```
alsuren, line 75: this will break `git bisect` I think: as soon as you check out an old version of the code, this test will start to fail.

alsuren (commented):
A bit more of an involved review this time, but it is quite a big refactor.

I mostly shied away from streaming from curl to tar the proper way, because it made errors harder to reason about. Be extra careful with this.

cargo-quickinstall/src/command_ext.rs
```diff
@@ -0,0 +1,111 @@
...
+
+impl CommandExt for Command {
+    fn formattable(&self) -> CommandFormattable<'_> {
+        CommandFormattable(self)
+    }
+
+    fn output_checked_status(&mut self) -> Result<Output, InstallError> {
+        self.output()
+            .map_err(InstallError::from)
+            .and_then(|output| check_status(self, output))
```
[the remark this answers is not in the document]
NobodyXu: > Could make output_checked_status() create a ChildWithCmd and call wait_with_output_checked_status() on it? That way, you can inline check_status() into it.

I tried but failed because `Command::spawn_with_cmd` calls `Command::spawn`, which by default inherits stdin, stdout and stderr.

While `Command::output_checked_status` calls `Command::output`, which by default read stdout and stderr into a pipe.

cargo-quickinstall/src/lib.rs
```diff
@@ -76,16 +84,21 @@ pub fn install_crate_curl(details: &CrateDetails, fallback: bool) -> Result<(),
...
-            e
-        }
-    })?;
-    Ok(parsed["versions"][0]["num"].clone().try_into().unwrap())
+        })?
+        .get_owned(&"versions")
+        .and_then(|value| value.get_owned(&0))
+        .and_then(|value| value.get_owned(&"num"))
+        .and_then(JsonValueExt::try_into_string)
+        .map_err(InstallError::from)
```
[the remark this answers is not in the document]
NobodyXu: Using `?` does seem to be leaner than `Result::and_then`.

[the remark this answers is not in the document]
NobodyXu: But the suggestion is incomplete and does not include `.try_into_string()?`

cargo-quickinstall/src/command_ext.rs
```diff
@@ -0,0 +1,111 @@
...
+        for arg in cmd.get_args() {
+            f.write_str(" ")?;
+            write_os_str(f, arg)?;
+        }
+
+        Ok(())
+    }
+}
+
+pub struct ChildWithCmd {
```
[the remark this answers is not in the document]
NobodyXu: Done

cargo-quickinstall/src/lib.rs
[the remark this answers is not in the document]
NobodyXu: Done

cargo-quickinstall/src/json_value_ext.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn unexpected(value: &JsonValue, expected: &str) -> Self {
+        JsonExtError(format!(
+            "Expecting {expected}, but found {}",
+            value.get_value_type()
+        ))
+    }
+}
+
+pub trait JsonValueExt {
+    fn get_owned(self, key: &dyn JsonKey) -> Result<JsonValue, JsonExtError>;
```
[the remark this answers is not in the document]
NobodyXu: Done

cargo-quickinstall/tests/integration_tests.rs
NobodyXu: Done

cargo-quickinstall/src/command_ext.rs
```diff
@@ -0,0 +1,111 @@
...
+
+impl fmt::Display for CommandFormattable<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let cmd = self.0;
+
+        write_os_str(f, cmd.get_program())?;
+
+        for arg in cmd.get_args() {
+            f.write_str(" ")?;
+            write_os_str(f, arg)?;
```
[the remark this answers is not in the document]
NobodyXu: I've added quotation for args that have ascii whitespace.

cargo-quickinstall/tests/integration_tests.rs
NobodyXu: Ok, so I've also added quotation for brackets by taking some code from [`shellecape`](https://docs.rs/shell-escape/latest/src/shell_escape/lib.rs.html#17-23).

cargo-quickinstall/tests/integration_tests.rs
NobodyXu: Fixed.
I've replaced this with a `git describe` to find the latest tag.

cargo-quickinstall/src/lib.rs
```diff
@@ -39,8 +47,8 @@ pub fn get_cargo_binstall_version() -> Option<String> {
 
 pub fn install_crate_curl(details: &CrateDetails, fallback: bool) -> Result<(), InstallError> {
     match download_tarball(&details.crate_name, &details.version, &details.target) {
-        Ok(tarball) => {
-            let tar_output = untar(tarball)?;
+        Ok(curl) => {
+            let tar_output = untar(curl)?;
```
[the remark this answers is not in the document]
NobodyXu: Thanks for pointing this out, I've fixed this.

cargo-quickinstall/src/lib.rs
NobodyXu: In `untar`, it will propagate error from `curl` first, which includes the 404 error, before propagating error from `untar`.

cargo-quickinstall/src/command_ext.rs
```diff
@@ -25,17 +25,43 @@ impl CommandExt for Command {
...
     }
 }
 
 pub struct CommandFormattable<'a>(&'a Command);
 
+fn needs_escape(s: &str) -> bool {
+    s.contains(|ch| !matches!(ch, 'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_' | '=' | '/' | ',' | '.' | '+'))
+}
+
 fn write_os_str(f: &mut fmt::Formatter<'_>, os_str: &OsStr) -> fmt::Result {
```
alsuren, line 41: we're no longer just writing the string, so write_os_str might not be the best name anymore

```suggestion
fn write_shell_arg_escaped(f: &mut fmt::Formatter<'_>, os_str: &OsStr) -> fmt::Result {
```

cargo-quickinstall/src/json_value_ext.rs
```diff
@@ -27,16 +27,16 @@ impl JsonExtError {
 }
 
 pub trait JsonValueExt {
-    fn get_owned(self, key: &dyn JsonKey) -> Result<JsonValue, JsonExtError>;
+    fn extract_from_value(self, key: impl JsonKey) -> Result<JsonValue, JsonExtError>;
```
alsuren, line 30: in my previous comment, I was complaining about JsonKey::get_owned. Please keep JsonValueExt::get_owned how it was

cargo-quickinstall/src/command_ext.rs
NobodyXu: Done

cargo-quickinstall/src/json_value_ext.rs
NobodyXu: Revertion done.

cargo-quickinstall/src/lib.rs
```diff
@@ -94,10 +101,10 @@ pub fn get_latest_version(crate_name: &str) -> Result<String, InstallError> {
...
         })?
-        .get_owned(&"versions")
-        .and_then(|value| value.get_owned(&0))
-        .and_then(|value| value.get_owned(&"num"))
-        .and_then(JsonValueExt::try_into_string)
+        .extract_from_value("versions")?
+        .extract_from_value(0)?
+        .extract_from_value("num")?
+        .try_into_string()
         .map_err(InstallError::from)
```
alsuren, line 108: (optional) if I'm not using Ok(err?) then I would normally write From::from (that what ? desugars into anyway)). This is one less thing to refactor if you change your error types later.

```suggestion
        .map_err(From::from)
```

alsuren (approved):
I've reviewed the diff since my last review and it looks good. I've not re-checked any of the threads that you marked as resolved.

Let's ship it.

When do you want to make the next release?

NobodyXu:
> When do you want to make the next release?

I think there's still something needs to be worked on:
 - Downloading binaries directly from cargo-binstall release, which I think provides more target support than quickinstall builds
 - improve error returned by `get_target_triple`.
    Currently if `host :` is not found, it is treated as a command failure while it's actually not.
    And it does not use the `CommandExt::output_checked_status`.

I suppose that another new reelase in 2-3ds?

PR #122, opened 2023-01-13
Tweak release build settings to reduce size and compilation time

NobodyXu:
Supersedes #118 

This PR is able to get both the binary size reduction as well as reduction in compilation time by using `-Oz`, stripping debug info and abort on panic.

It was able to reduce size from 700K => 476K while reducing compilation time on my M1 from 1s => 0.75s.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,3 +2,8 @@
 members = [
     "cargo-quickinstall",
 ]
+
+[profile.release]
+opt-level = "z"
+panic = "abort"
+strip = "symbols"


NobodyXu:
Compare to #118 , this PR disables lto and uses the default `codegen-units`, which generates binary size a little larger than that PR (about 50K) but the reduction in compilation time is well worth it.

Disabling the lto plus `codegen-units` tweak actually improves compilation time.

NobodyXu:
The CI also runs pretty fast with no regression.

alsuren (approved):
Sounds good as long as we're comfortable with panic=abort.

I will try to sit down and review your other pr over the weekend.

I'm idly wondering if we could somehow prove that we don't panic - add https://github.com/dtolnay/no-panic as a dev dependency, and add `#[cfg_attr(test, no_panic)] in the relevant place.

NobodyXu:
> Sounds good as long as we're comfortable with panic=abort.

In principle `cargo-quickinstall` should not panic at all.

> I'm idly wondering if we could somehow prove that we don't panic - add https://github.com/dtolnay/no-panic as a dev dependency, and add `#[cfg_attr(test, no_panic)] in the relevant place.

That actually relies on compiler optimization to remove unreachable `panic!`, so it might need the `profile.dev.opt-level` to set to `3`.

Also, #121 has removed most, if not all (including the implicit one, e.g. accessing `JsonValue` using `Index` trait) `unwrap()` and replaced them with proper error testing.

AFAIK, there's only one `unwrap` now in #121 when taking `curl.stdout().take().unwrap()`, so it should be fine.

PR #123, opened 2023-01-13
milo/support aarch64 unknown gnu

Milo123459:
This allows `cargo-quickbuild` to build for `aarch64-unknown-linux-gnu`.

### DIFF
--- a/build-version.sh
+++ b/build-version.sh
@@ -30,6 +30,20 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
 
     CARGO_BIN_DIR="${TEMPDIR}/cargo/bin"
     CRATES2_JSON_PATH="${TEMPDIR}/cargo/.crates2.json"
+elif [ "$TARGET_ARCH" == "aarch64-unknown-linux-gnu" ]; then
+    mkdir -p zigfolder
+    curl "$(curl -q https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//')" | tar -xJ -C zigfolder --strip-components 1
+
+    export PATH="$PWD/zigfolder:$PATH"
+    rustup target add "$TARGET_ARCH"
+    if ! [ -f "$HOME/.cargo/config" ]; then
+        echo "[target.aarch64-unknown-linux-gnu]" >>~/.cargo/config
+        echo "linker = \"$PWD/zig-aarch64.sh\"" >>~/.cargo/config
+    fi
+
+    CARGO_PROFILE_RELEASE_CODEGEN_UNITS="1" CARGO_PROFILE_RELEASE_LTO="fat" OPENSSL_STATIC=1 CC=./zig-aarch64.sh cargo install "$CRATE" --version "$VERSION" --target "$TARGET_ARCH"
+    CARGO_BIN_DIR=~/.cargo/bin
+    CRATES2_JSON_PATH=~/.cargo/.crates2.json
 else
     rustup target add "$TARGET_ARCH"
     CARGO_PROFILE_RELEASE_CODEGEN_UNITS="1" CARGO_PROFILE_RELEASE_LTO="fat" OPENSSL_STATIC=1 cargo install "$CRATE" --version "$VERSION" --target "$TARGET_ARCH"
--- a/zig-aarch64.sh
+++ b/zig-aarch64.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+# shellcheck disable=SC2068
+exec zig cc -target aarch64-linux-gnu $@
--- a/.github/workflows/selfbuild.yml
+++ b/.github/workflows/selfbuild.yml
@@ -23,6 +23,8 @@ jobs:
             os: ubuntu-20.04
           - target_arch: x86_64-unknown-linux-musl
             os: ubuntu-20.04
+          - target_arch: aarch64-unknown-linux-gnu
+            os: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
         with:
--- a/trigger-package-build.sh
+++ b/trigger-package-build.sh
@@ -14,6 +14,8 @@ get_build_os() {
         echo "ubuntu-20.04"
     elif [[ "$1" == "x86_64-pc-windows-msvc" ]]; then
         echo "windows-latest"
+    elif [[ "$1" == "aarch64-unknown-linux-gnu" ]]; then
+        echo "ubuntu-20.04"
     else
         echo "Unrecognised build OS: $1"
         exit 1
@@ -46,7 +48,7 @@ main() {
         git config user.name "trigger-package-build.sh"
     fi
 
-    TARGET_ARCHES="${TARGET_ARCHES:-${TARGET_ARCH:-x86_64-pc-windows-msvc x86_64-apple-darwin aarch64-apple-darwin x86_64-unknown-linux-gnu x86_64-unknown-linux-musl}}"
+    TARGET_ARCHES="${TARGET_ARCHES:-${TARGET_ARCH:-x86_64-pc-windows-msvc x86_64-apple-darwin aarch64-apple-darwin x86_64-unknown-linux-gnu x86_64-unknown-linux-musl aarch64-unknown-linux-gnu}}"
 
     if [ ! -d "${TEMPDIR:-}" ]; then
         TEMPDIR="$(mktemp -d)"


Milo123459:
cc @NobodyXu - wasn't sure how to test. this should in theory work, but needs to be tested in CI.

Milo123459:
lmk if you have any ideas on how i can fix this because i have no idea how

build-version.sh
```diff
@@ -30,6 +30,20 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
 
     CARGO_BIN_DIR="${TEMPDIR}/cargo/bin"
     CRATES2_JSON_PATH="${TEMPDIR}/cargo/.crates2.json"
+elif [ "$TARGET_ARCH" == "aarch64-unknown-linux-gnu" ]; then
+    wget $(curl https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//') -O zig
```
NobodyXu, line 34: The error from the [`shellcheck`](https://github.com/cargo-bins/cargo-quickinstall/actions/runs/3915594296/jobs/6693920222) indicates that you need to quote this to prevent word splitting:

```suggestion
    wget "$(curl https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//')" -O zig
```

zig-aarch64.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+zig cc -target aarch64-unknown-linux-gnu $@
```
NobodyXu, line 2:
```suggestion
# shellcheck disable=SC2068
zig cc -target aarch64-unknown-linux-gnu $@
```

to disable the [`shellcheck`](https://github.com/cargo-bins/cargo-quickinstall/actions/runs/3915594296/jobs/6693920222#step:3:12) here.

NobodyXu (changes requested):
For the linter error:

build-version.sh
NobodyXu: Note that I use `-q` here to disable the progress bar, which takes up a lot of space in the CI log, makes the debugging harder.

build-version.sh
```diff
@@ -30,6 +30,20 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
...
+    mkdir zigfolder
+    tar -xf ./zig -C zigfolder --strip-components 1
+    rm zig
+    sudo mv zigfolder/zig /usr/local/bin
+    rustup target add "$TARGET_ARCH"
+    echo "[target.aarch64-unknown-linux-gnu]" >> ~/.cargo/config
+    echo '["linker = "./zig-aarch64.sh"]' >> ~/.cargo.config
+
+    CARGO_PROFILE_RELEASE_CODEGEN_UNITS="1" CARGO_PROFILE_RELEASE_LTO="fat" OPENSSL_STATIC=1 cargo install "$CRATE" CC=./zig-aarch64.sh --version "$VERSION" --target "$TARGET_ARCH"
```
NobodyXu, line 43: According to the [CI](https://github.com/cargo-bins/cargo-quickinstall/actions/runs/3915594303/jobs/6693920659#step:3:960), cargo still uses `ld`.

build-version.sh
```diff
@@ -30,6 +30,20 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
...
     CRATES2_JSON_PATH="${TEMPDIR}/cargo/.crates2.json"
+elif [ "$TARGET_ARCH" == "aarch64-unknown-linux-gnu" ]; then
+    wget $(curl https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//') -O zig
+    mkdir zigfolder
+    tar -xf ./zig -C zigfolder --strip-components 1
+    rm zig
+    sudo mv zigfolder/zig /usr/local/bin
+    rustup target add "$TARGET_ARCH"
+    echo "[target.aarch64-unknown-linux-gnu]" >> ~/.cargo/config
+    echo '["linker = "./zig-aarch64.sh"]' >> ~/.cargo.config
```
NobodyXu, line 41: I think we should instead maintain a new file `cargo-config` since using `echo` to create a file is basically unreadable.

The file should look like this according to [this SSO](https://stackoverflow.com/questions/62371337/rust-cargo-toml-specify-custom-path-for-c-linker-and-compiler):

```
[target.aarch64-unknown-linux-gnu]
linker = "./zig-aarch64.sh"
ar = "./zig-aarch64.shr"
```

NobodyXu (changes requested):

build-version.sh
Milo123459: weird...

build-version.sh
Milo123459: hmm, i think for now it is fine.

Milo123459: i was going off of https://actually.fyi/posts/zig-makes-rust-cross-compilation-just-work/

NobodyXu: I think that's partially because:

```
echo '["linker = "./zig-aarch64.sh"]'
```

prints:

```
["linker = "./zig-aarch64.sh"]
```

which is not what cargo expected.

Milo123459: I updated it in a recent commit but it's still broken

NobodyXu:
Could you be hitting https://github.com/ziglang/zig/issues/4911 ?

Milo123459:
I think so.

Milo123459:
getting a new error this time. cc @NobodyXu - any ideas?

Milo123459:
right - now im trying to make it use  the NDK version of ld. currently, when i move ld to /usr/bin, the shell doesn't pick up that there is something in that directory called `ld` and now errors out because it cannot find it. it turns out that the `sudo mv` isn't actually moving the file, i think

Milo123459:
cc @alsuren - any ideas

zig-aarch64.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+# shellcheck disable=SC2068
+zig cc -target aarch64-linux-gnu $@
```
alsuren, line 3:
```suggestion
exec zig cc -target aarch64-linux-gnu $@
```

no need to keep the shell process around in the process tree while cc runs

build-version.sh
```diff
@@ -30,6 +30,22 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
 
     CARGO_BIN_DIR="${TEMPDIR}/cargo/bin"
     CRATES2_JSON_PATH="${TEMPDIR}/cargo/.crates2.json"
+elif [ "$TARGET_ARCH" == "aarch64-unknown-linux-gnu" ]; then
+    wget "$(curl -q https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//')" -O zig -q
+    mkdir -p zigfolder
+    tar -xf ./zig -C zigfolder --strip-components 1
+    rm zig
+    sudo mv zigfolder/zig /usr/local/bin/zig
+    sudo mv zigfolder/lib /usr/local/bin/lib
```
alsuren, line 39: * probably just pipe the tarball straight into tar
* can just put the zig dir first in your $PATH rather than copying stuff into /usr/local/bin?


(I'm not sure what `/usr/local/bin/lib` is supposed to be, but it feels wrong)
```suggestion
    mkdir -p zigfolder
    curl "$(curl -q https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//')"  | tar -x -C zigfolder --strip-components 1

    export PATH="$PWD/zigfolder:$PATH"
```

alsuren (approved):
Please give your squashed commit a better name, and consider my suggestions (it may be that you already tried them)

I would quite like to try one of these binaries in docker on my m1 mac. Let's ship it and get it building stuff :D

build-version.sh
Milo123459: we need /lib or else it'll error with zig not being able to find some lib folder, so yeah the way it was is how we have to do it

Milo123459:
i can't merge it because there are some review changes that haven't been addressed (even though they have). @alsuren could you merge it for me? thanks

alsuren:
The "request changes" review is blocking the merge. I tend to avoid using that button because it feels impolite (I always tend to leave feedback as "comment"). I hadn't realised that it also blocks merging in the presence of other approvers :-(.

If you want to smoke test it before we merge to main, push it to the `actions` branch and it will kick off a one-off cronjob using your code.

Milo123459:
alright, i'll push it! fingers crossed

Milo123459:
seems to have worked! should i delete the `actions` branch?

Milo123459:
nvm, it's failing because it's using a weird version of the branch causing it to not have access to a file. in practice, this won't happen if this is merged and the cron just runs.

ps @alsuren - can you not override the requested changes thing and merge it?

build-version.sh
alsuren: The error you got from CI was about my broken tar command (I left out the -J)


```suggestion
    mkdir -p zigfolder
    curl "$(curl -q https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//')"  | tar -xJ -C zigfolder --strip-components 1

    export PATH="$PWD/zigfolder:$PATH"
```

The only reason you need to move the lib folder is that the zig binary is looking relative to itself for its lib foder. If you don't move the zig binary (and instead add it to $PATH) then I would expect this to work just fine.

alsuren:
> ps @alsuren - can you not override the requested changes thing and merge it?

Oh hey: there is a "dismiss review" item in the menu.

I still think that https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4033938309/jobs/6934774496#step:4:89  (
```
error: linker `/home/runner/work/cargo-quickinstall/cargo-quickinstall/zig-aarch64.sh` not found
```
) is a problem that we're also going to see on main. Let me have a quick dig about.

build-version.sh
```diff
@@ -30,6 +30,20 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
...
     CRATES2_JSON_PATH="${TEMPDIR}/cargo/.crates2.json"
+elif [ "$TARGET_ARCH" == "aarch64-unknown-linux-gnu" ]; then
+    mkdir -p zigfolder
+    curl "$(curl -q https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//')" | tar -xJ -C zigfolder --strip-components 1
+
+    export PATH="$PWD/zigfolder:$PATH"
+    rustup target add "$TARGET_ARCH"
+    if ! [ -f "$HOME/.cargo/config" ]; then
+        echo "[target.aarch64-unknown-linux-gnu]" >>~/.cargo/config
+        echo 'linker = "/home/runner/work/cargo-quickinstall/cargo-quickinstall/zig-aarch64.sh"' >>~/.cargo/config
```
alsuren, line 41: I managed to make https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4034113715/jobs/6935051984 pass. I think the thing that did the trick was this:

```suggestion
        echo "linker = \"$PWD/zig-aarch64.sh\"" >>~/.cargo/config
```

You can also try cherry-picking the thing that makes our selftest stuff checkout the repo in the same place as the builder does. Annoyingly it is /home/runner/work/cargo-quickinstall/cargo-quickinstall/cargo-quickinstall/ :( Might be a thing for another PR though.

alsuren:
I just tested it out in arm docker, and `cargo-quickinstall cargo-quickinstall` now works (once I installed `curl`). Good work, and thanks for sticking with it.

Glancing at https://github.com/cargo-bins/cargo-quickinstall/pull/123/commits still doesn't tell me anything about what this PR is about, so it's likely to cause problems for anyone doing archaeology later (although they will have enough problems unpicking my commits, so they might not notice yours).

Otherwise, you have an approval and I *think* you have merge access, so ship it.

Now it's well past my bedtime. I might do some more quickinstall hacking tomorrow though.

Milo123459:
Woo! I'll squash my commits again. Just testing the actions branch then I'll give it a merge!

build-version.sh
```diff
@@ -30,6 +30,20 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
...
+    export PATH="$PWD/zigfolder:$PATH"
+    rustup target add "$TARGET_ARCH"
+    if ! [ -f "$HOME/.cargo/config" ]; then
+        echo "[target.aarch64-unknown-linux-gnu]" >>~/.cargo/config
+        echo "linker = \"$PWD/zig-aarch64.sh\"" >>~/.cargo/config
+    fi
+
+    CARGO_PROFILE_RELEASE_CODEGEN_UNITS="1" CARGO_PROFILE_RELEASE_LTO="fat" OPENSSL_STATIC=1 CC=./zig-aarch64.sh cargo install "$CRATE" --version "$VERSION" --target "$TARGET_ARCH"
+    CARGO_BIN_DIR=~/.cargo/bin
+    CRATES2_JSON_PATH=~/.cargo/.crates2.json
```
NobodyXu, line 46: Excellent job @Milo123459 !

Looks like the issue I linked is already fixed, that's great because I also want to use zig-cc in cargo-binstall.

@Milo123459 One thing that I would you like to fix here is to use `cargo-auditable` to build the binary, just as other branches do.

You can also rewrite the other branches (the musl one) to use zig-cc and then we can also support `aarch64-unknown-linux-musl`.

Milo123459:
Seeing some errors @alsuren 😬 https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4034450545/jobs/6935671373#step:5:101

PS: I ran `cargo quickinstall cargo-binstall` on my Raspberry Pi, that has the aarch64-unknown-linux-gnu target, but a workflow run was never created for it.

NobodyXu:
This error is caused by my PR  #140, will fix it

PR #125, opened 2023-01-14
Fix `test_get_latest_version`: Use `git-tag` instead of `git-describe`

NobodyXu:
The former works more reliably, including on a shallow repository with tags fetched.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,10 +40,8 @@ jobs:
       - name: Fetch source
         uses: actions/checkout@v3
 
-      - name: Fetch tags and origin/main required for testing
-        run: |
-          git fetch --tags
-          git fetch origin main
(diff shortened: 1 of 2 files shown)


alsuren (approved):
This still feels fragile (git tag might not always line up with crates.io release), but I can't think of a better way to do it without adding even more complexity (mocking out curl), so let's go with this approach and see how long it lasts. This is always the way with integration tests, I suppose.

NobodyXu:
> This still feels fragile (git tag might not always line up with crates.io release), but I can't think of a better way to do it without adding even more complexity (mocking out curl), so let's go with this approach and see how long it lasts. This is always the way with integration tests, I suppose.

The most robust way is to use `crates_io_api`, but that would significantly increase compilation time for the tests on CI.

PR #132, opened 2023-01-16
Add support for batch installation mode

NobodyXu:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,3 +72,16 @@ jobs:
         run: |
           cargo r -- cargo-quickinstall
           cargo quickinstall -V
+
+      - name: Test batch installation with binstall
+        run: |
+          cargo r -- cargo-quickinstall cargo-nextest
+          cargo quickinstall -V
+          cargo nextest -V
+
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -72,3 +72,16 @@ jobs:
         run: |
           cargo r -- cargo-quickinstall
           cargo quickinstall -V
+
+      - name: Test batch installation with binstall
+        run: |
+          cargo r -- cargo-quickinstall cargo-nextest
```
alsuren, line 78: this should fail on windows.... except it doesn't because we `exec` to run binstall, so we're no longer running? or does it noop because it's already installed?

alsuren (approved):
I've only skimmed it. I trust you to get this right.

.github/workflows/test.yml
NobodyXu: > this should fail on windows....

Why shall this fail though?

> except it doesn't because we exec to run binstall

We only do exec on unix.

> or does it noop because it's already installed?

Yes, `cargo-binstall` checks local manifest and do nothing if it's already installed.

PR #134, opened 2023-01-17
Add `cargo-auditable` to `popular-crates.txt`

NobodyXu:
I'd like quickinstall to provide pre-built artifacts for `cargo-auditable` so that we can use it in cargo-binstall and cargo-quickinstall.

### DIFF
--- a/popular-crates.txt
+++ b/popular-crates.txt
@@ -20,6 +20,7 @@
 # The first empty line is important. I can't remember why.
 
 cargo-quickinstall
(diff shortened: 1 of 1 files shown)


alsuren (approved):
I'm not going to be super-responsive for the next few weeks. Feel free to merge things like this without a review.

Also, if I don't get back to you on bigger things quickly, re-review them yourself and merge them.

NobodyXu:
Got it

PR #136, opened 2023-01-18
Fallback to `guess_host_triple` if `rustc -vV` failed

NobodyXu:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,7 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 name = "cargo-quickinstall"
 version = "0.2.7"
 dependencies = [
+ "guess_host_triple",
  "home",
  "mktemp",
  "pico-args",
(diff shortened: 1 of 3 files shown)


alsuren:
What are the circumstances that actually cause rustc -vV to fail? This feels like an edge case that isn't worth 30 seconds of build time - https://github.com/cargo-bins/cargo-quickinstall/pull/149#issuecomment-1407688462

NobodyXu:
> What are the circumstances that actually cause rustc -vV to fail? This feels like an edge case that isn't worth 30 seconds of build time - https://github.com/cargo-bins/cargo-quickinstall/pull/149#issuecomment-1407688462

Well, it fails when rustc isn't present.

If we can assume rustc always prrsent, then we can revert this.

alsuren:
I don't think we provide any instructions for installing cargo-quickinstall without using cargo. We also install things into a place that would only be in your $PATH if you are using rust, and if something isn't installed then we fallback to using cargo install.

It seems reasonable to assume that cargo and rustc are installed. It might be worth going around and adding nice error messages for when our runtime dependencies aren't available.

PR #140, opened 2023-01-18
Use `cargo-auditable` to build the binaries

### LINKED ISSUE #133: Use `cargo-auditable` to build binaries
Using [`cargo-auditable`](https://github.com/rust-secure-code/cargo-auditable) will produce binaries along with an embedded dependencies list, which makes it possible for tools like `cargo-audit` to recover the information and scan for vulnerabilities.

NobodyXu:
Fixed #133

so that it contains version information that can be extracted by tools such as `cargo-audit` to check for vulnerabilities.

### DIFF
--- a/.github/workflows/selfbuild.yml
+++ b/.github/workflows/selfbuild.yml
@@ -27,6 +27,11 @@ jobs:
       - uses: actions/checkout@v3
         with:
           persist-credentials: false
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-auditable
+
       - name: Build Thyself
         env:
           TARGET_ARCH: ${{ matrix.target_arch }}
--- a/build-version.sh
+++ b/build-version.sh
@@ -25,7 +25,7 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
     # Compiling against musl libc is failing despite installing the musl-tools
     # deb.  Falling back to Rust's Alpine container whose default target
     # is x86_64-unknown-linux-musl.
-    podman run --name=official-alpine-rust docker.io/library/rust:alpine sh -c "apk update && apk add build-base curl; curl --location --silent --show-error --fail https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-$TARGET_ARCH.tgz | tar -xzvvf -; ./cargo-binstall binstall cargo-auditable; CARGO_PROFILE_RELEASE_CODEGEN_UNITS=\"1\" CARGO_PROFILE_RELEASE_LTO=\"fat\" OPENSSL_STATIC=1 cargo auditable install $CRATE --version $VERSION"
+    podman run --name=official-alpine-rust docker.io/library/rust:alpine sh -c "apk update && apk add build-base curl; curl --location --silent --show-error --fail https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-$TARGET_ARCH.tgz | tar -xzvvf -; ./cargo-binstall binstall -y cargo-auditable; CARGO_PROFILE_RELEASE_CODEGEN_UNITS=\"1\" CARGO_PROFILE_RELEASE_LTO=\"fat\" OPENSSL_STATIC=1 cargo auditable install $CRATE --version $VERSION"
     podman cp official-alpine-rust:/usr/local/cargo "${TEMPDIR}/"
 
     CARGO_BIN_DIR="${TEMPDIR}/cargo/bin"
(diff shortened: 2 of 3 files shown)


Milo123459 (commented):
Where is the audit information stored?

NobodyXu:
> Where is the audit information stored?

It's stored directly in the binary.

Checkout the [`cargo-auditable`](https://github.com/rust-secure-code/cargo-auditable) gh repo, it contains more information on this.

Milo123459:
Hmm, just a quick idea: when a binary is installed from the quickinstall tags, could we make it say something about the vulnerabilities, if there are any, after install?

NobodyXu:
> Hmm, just a quick idea: when a binary is installed from the quickinstall tags, could we make it say something about the vulnerabilities, if there are any, after install?

That can be possible, but that would require `cargo-quickinstall` to download `cargo-audit` to run the checks.

Milo123459:
Oh.

NobodyXu:
@alsuren Pinging as this has been stale for roughly one week.

build-version.sh
```diff
@@ -25,14 +25,14 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
...
 
     CARGO_BIN_DIR="${TEMPDIR}/cargo/bin"
     CRATES2_JSON_PATH="${TEMPDIR}/cargo/.crates2.json"
 else
     rustup target add "$TARGET_ARCH"
-    CARGO_PROFILE_RELEASE_CODEGEN_UNITS="1" CARGO_PROFILE_RELEASE_LTO="fat" OPENSSL_STATIC=1 cargo install "$CRATE" --version "$VERSION" --target "$TARGET_ARCH"
+    CARGO_PROFILE_RELEASE_CODEGEN_UNITS="1" CARGO_PROFILE_RELEASE_LTO="fat" OPENSSL_STATIC=1 cargo auditable install "$CRATE" --version "$VERSION" --target "$TARGET_ARCH"
 
     CARGO_BIN_DIR=~/.cargo/bin
     CRATES2_JSON_PATH=~/.cargo/.crates2.json
```
alsuren, line 38: From memory, we parse ~/.cargo/.crates2.json to see which binaries have been installed, and then use that for the list of files to include in the tarball. Is there any way to install cargo-auditable somewhere else in the path, and skip writing to ~/.cargo/.crates2.json?

alsuren: > install cargo-auditable somewhere else in the path

specfically so that `cargo auditable install cargo-auditable` won't fail if we are trying to build cargo-auditable.

NobodyXu: Yes, we can configure cargo-binstall to skip that

.github/workflows/selfbuild.yml
```diff
@@ -27,6 +27,11 @@ jobs:
       - uses: actions/checkout@v3
         with:
           persist-credentials: false
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-auditable
+
       - name: Build Thyself
```
alsuren, line 35: it's occurred to me that the "Build Thyself" step chould probably assert that `"${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"` has been created and that

[[ "$(tar --list -f "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz")" == "cargo-quickinstall" ]]

alsuren: I suppose it could also upload the tarball so you can fetch it and check manually?

NobodyXu: That's doable, but I feel perhaps it's better as a separare PR?

alsuren (approved):
On this kind of PR, I would probably just ship it and then check that the resulting binaries do the right thing on all arches once it's built some packages.

If something is wrong (like it's shipping cargo-auditable with all other packages) then just delete the tags and hope nobody noticed ;-). It's okay to be wrong as long as you've got a plan to recover.

A large part of software engineering is optimising cycle time: "how many times you can get it wrong in a minute without anyone noticing?". The more quickly you can iterate and get it wrong, the more quickly you will learn, and the more quickly you will arrive at the right answer. Obviously a good compiler will help with this, and good tests in CI goes a long way, but sometimes you just have to get things wrong in production and be ready to clean up after yourself ;-).

If you want to smoke-test builder-related things without having to push to main (which triggers builds every hour), you can instead push to the `actions` branch (which is magic and triggers a build every time you push to it). I suppose it's not much use if there's nothing in the queue of packages to build, but sometimes it's useful.

Sorry for leaving you hanging so long on this one. I'm mostly sorry for not leaving the project with enough automated checks to let people be confident at trying things.

NobodyXu:
No worries!
I will fix the cargo-auditable by skipping updating manifest then merge it.

build-version.sh
NobodyXu: I've fixed this by removing `.crates2.json` and I just realized that `cargo-binstall` does not update `.crates2.json` anyway.

Issue #126, opened 2023-01-14
feedback from testing ready for a release


alsuren:
We're about to do a release, so I thought I'd do some manual testing.

I will add to the list as I find more stuff. Feel free to add stuff as well, and/or check things off.

- [ ] I've just realised that the output now contains a warning that doesn't make much sense if you don't know that we're using binstall:

```
$ cargo quickinstall bat
 INFO resolve: Resolving package: 'bat'
 WARN The package bat v0.22.1 will be downloaded from third-party source QuickInstall
 INFO This will install the following binaries:
 INFO   - bat (bat -> /Users/alsuren/.cargo/bin/bat-v0.22.1)
 INFO And create (or update) the following symlinks:
 INFO   - bat (/Users/alsuren/.cargo/bin/bat -> bat-v0.22.1)
 INFO Installing binaries...
 INFO Done in 10.728787291s
```

We should probably either find a way to suppress that warning or tell the user that we're calling out to cargo-binstall to install the package ("use 'cargo quickinstall --no-binstall' to fetch packages directly if you don't need 'cargo uninstall' to work)

As an aside, I don't normally write the CamelCase `QuickInstall`. I tend to write it as the command is written, in lowercase - `cargo quickinstall` `cargo-quickinstall` or just `quickinstall`, and sometimes use backticks if I'm putting it in a title or the start of a sentence.

- [ ] 10 seconds is a long time. We used to manage it in less than 1 second.

It's a bit sad that `cargo install cargo-quickinstall` is now faster than `cargo quickinstall cargo-quickinstall` (actually `cargo uninstall cargo-quickinstall && cargo run -- quickinstall cargo-quickinstall` - see below about --force)

In the grand scheme of things you will probably only be using cargo-quickinstall if you know you're installing a huge package, so maybe it's fine.

- [ ] `cargo quickinstall cargo-quickinstall` (from above) actually no-ops and prints `cargo-quickinstall v0.2.7 is already installed, use --force to override`. We don't support a --force flag though, because we previously always untarred on top of whatever was there.

- [ ] The stats curl call manages to print its output after quickinstall has exited and bash has already rendered its prompt for the next command

``` 
alsuren@brute:~/src/cargo-quickinstall main via 🐍 v3.10.9 (.venv) via 🦀 v1.65.0 via ⍱ 
$ time cargo quickinstall bat --no-binstall
Installed bat 0.22.1 to ~/.cargo/bin:
[...]

PR #149, opened 2023-01-29
Fixes for my release feedback notes

alsuren:
This addresses all of the things I noticed in https://github.com/cargo-bins/cargo-quickinstall/issues/126

I decided to disable versioned symlinks because I find them annoying (they break my tab completion) and they're not something that `cargo install` does natively.

I decided to make searching for upstream packages opt-in because they're slow. Disabling that strategy brings the time for `cargo quickinstall bat --force` down from 10 seconds to 2 seconds. Still not as good as `cargo quickinstall bat --force --no-binstall` (1 second), but good enough for me. I also quite like the simplicity of the explanation "`cargo quickinstall $package` will always give you something equivalent to `cargo install $package`". I have allowed users to opt-in with the `--try-upstream` flag. There might be a better name than `--try-upstream`.

I decided that --no-binstall was a perfectly fine name for the command-line flag, so I kept it.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,21 +61,21 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Test dry-run
-        run: cargo r -- --dry-run cargo-quickinstall
+        run: cargo run -- --dry-run cargo-quickinstall
 
       - name: Test installing and using binstall
         run: |
-          cargo r -- cargo-quickinstall
+          cargo run -- cargo-quickinstall
           cargo quickinstall -V
 
       - name: Test using binstall with it already installed
         run: |
-          cargo r -- cargo-quickinstall
+          cargo run -- cargo-quickinstall
           cargo quickinstall -V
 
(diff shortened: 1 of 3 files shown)


NobodyXu (approved):

NobodyXu:
LGTM, but the e2etest CI takes 6m since it built `cargo-nextest` from source due to disabling lookup from upstream.
Could change to another crate.

alsuren:
I am surprised that we don't have a build of nextest, but it builds fine in our CI. Have they made a recent release or something?

alsuren:
https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4037354021/jobs/6940594743#step:3:31

>    Compiling cargo-quickinstall v0.2.7 (/home/runner/work/cargo-quickinstall/cargo-quickinstall/cargo-quickinstall)
>    Finished dev [unoptimized + debuginfo] target(s) in 37.45s

Jesus - that's a long time.

PR #150, opened 2023-01-29
fix build for aarch64 with a bit of test driven development

alsuren:


.github/workflows/selfbuild.yml
```diff
@@ -52,3 +52,49 @@ jobs:
...
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-auditable
+
+      # FIXME: find a package that needs a working CC but doesn't take quite so long to build.
+      - name: Build cargo-deny
```
alsuren, line 88: @Milo123459 I am quite tempted to:
* do a full build sweep with the current broken CC
* do a full build sweep with this PR merged in
* get a list of packages that are broken under the broken CC but work under the fixed CC, and put the quickest one in here

Alternatively, we could just not bother merging a regression test for this.

alsuren: I have kicked off a job to attempt builds using what's in main every 5 minutes

`while sleep $((5*60)); do TARGET_ARCHES=aarch64-unknown-linux-gnu ./trigger-package-build.sh ; done`

Quite a lot of packages build just fine, even if CC is not available.

Milo123459: hmm, so why is cargo-deny failing? can we make it work?

alsuren: I'm now running the same loop on this branch, to see what builds with this patch applied.

> hmm, so why is cargo-deny failing? can we make it work?

Some packages just fail to build on some architectures. That's just life. We have code to automatically disable packages per-arch if they fail too many times, to avoid wasting CI resources. I've switched it to one that *does* build on most arches and takes about 2 minutes to build. Still worth trying to find a quicker package, but should be good enough to merge.

Milo123459 (approved):

.github/workflows/selfbuild.yml
alsuren: .... hrm... tin-summer's build failure was nothing to do with CC - https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4036562588/jobs/6939242828 . Maybe I should look more closely to find a good package for this CI check.

Milo123459: im rerunning it

Milo123459: yeah it worked

alsuren: oops: I just cancelled your rerun of https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4036562588 because it looked like it had been running for 2 hours or something

Milo123459: on another note: do you know why musl gives so many linker errors? idk of anything we can do really

alsuren: Honestly I don't use musl with rust if I can possibly avoid it (its allocator is unacceptably slow, and debian-slim is good enough for making small docker images these days), so I've not looked into it. Feel free to trawl the logs looking for patterns though. There might be a common theme that's easy to fix.

PR #154, opened 2023-01-29
(cargo-release) version 0.2.8

alsuren:


alsuren:
I'm deeply regretting that `cargo-deny` is the package I picked, because it takes ~10 minutes on some architectures. Maybe we actually *don't* want to check that CC is set correctly, if this is what it does to our CI times :/ .

alsuren:
I jumped the gun on making my PR because some arches hadn't finished building yet. Once the cargo-deny builds had finished, I hit "rerun" on the jobs that failed/auto-cancelled, and it was fine.

PR #160, opened 2023-02-25
Switch to serde_json + serde + wa-serde-derive

NobodyXu:


NobodyXu:
I will test its performance tomorrow, but looking at the Self-Build CI, the compilation time looks ok to me and building this on my M1 took 5s.

NobodyXu:
Looking at the runtime performance on the GHA, which has faster internet connection than any of us ever have, it looks like it doesn't yield much benefit.

NobodyXu:
Closed as it doesn't yield any benefit.

PR #163, opened 2023-03-03
New release/tag format: $CRATE-$VERSION

NobodyXu:

### DIFF
--- a/.github/workflows/build-package.yml.template
+++ b/.github/workflows/build-package.yml.template
@@ -68,6 +68,9 @@ jobs:
     name: Upload
     needs: build-popular-package
     runs-on: ubuntu-20.04
+    defaults:
+      run:
+        shell: bash
     steps:
       - name: Checkout trigger commit
         uses: actions/checkout@v3
@@ -109,16 +112,34 @@ jobs:
       - name: Tag release
         run: |
           (
+            set -euxo pipefail
             cd trigger
-            git tag $CRATE-$VERSION-$TARGET_ARCH
-            git push origin $CRATE-$VERSION-$TARGET_ARCH
+            if ! ../cargo-quickinstall/create_and_upload_tag.sh $CRATE-$VERSION; then
+                echo "$CRATE-$VERSION tag already exists"
+            fi
+            if ! ../cargo-quickinstall/create_and_upload_tag.sh $CRATE-$VERSION-$TARGET_ARCH; then
+                echo "$CRATE-$VERSION-$TARGET_ARCH tag already exists"
+                echo 'SKIP_OLD_FORMAT=1' >>$GITHUB_ENV
+            fi
           )
 
-      - name: Releasing assets
-        uses: softprops/action-gh-release@v1
+      - name: Releasing assets in old release format
+        uses: svenstaro/upload-release-action@v2
+        if: "! env.SKIP_OLD_FORMAT"
         with:
-          tag_name: $CRATE-$VERSION-$TARGET_ARCH
-          files: |
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          tag: $CRATE-$VERSION-$TARGET_ARCH
+          release_name: $CRATE-$VERSION-$TARGET_ARCH
+          body: build $CRATE@$VERSION on $TARGET_ARCH
+          file: |
             ${{github.workspace}}/built.d/$CRATE-$VERSION-$TARGET_ARCH.tar.gz
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Releasing assets in new release format
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          tag: $CRATE-$VERSION
+          release_name: $CRATE-$VERSION
+          body: build $CRATE@$VERSION
+          file: |
+            ${{github.workspace}}/built.d/$CRATE-$VERSION-$TARGET_ARCH.tar.gz
--- a/build-version.sh
+++ b/build-version.sh
@@ -16,20 +16,28 @@ curl_slowly() {
     sleep 1 && curl --user-agent "cargo-quickinstall build pipeline (alsuren@gmail.com)" "$@"
 }
 
-if [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -I --output /dev/null "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}-${TARGET_ARCH}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
+if [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -I --output /dev/null "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
     echo "${CRATE}/${VERSION}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz already uploaded. Skipping."
+    exit 0
+elif [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -L --output "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}-${TARGET_ARCH}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
+    echo "copy-pasting from the old location rather than building from scratch" >&2
+    echo "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"
     exit 0
 fi
 
 if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
+    CARGO_ROOT=$(mktemp -d 2>/dev/null || mktemp -d -t 'cargo-root')
+
     # Compiling against musl libc is failing despite installing the musl-tools
     # deb.  Falling back to Rust's Alpine container whose default target
     # is x86_64-unknown-linux-musl.
-    podman run --name=official-alpine-rust docker.io/library/rust:alpine sh -c "set -euxo pipefail; apk update && apk add build-base curl; curl --location --silent --show-error --fail https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-$TARGET_ARCH.tgz | tar -xzvvf -; ./cargo-binstall binstall -y cargo-auditable; rm -f \$CARGO_HOME/.crates.toml \$CARGO_HOME/.crates2.json; CARGO_PROFILE_RELEASE_CODEGEN_UNITS=\"1\" CARGO_PROFILE_RELEASE_LTO=\"fat\" OPENSSL_STATIC=1 cargo auditable install $CRATE --version $VERSION"
-    podman cp official-alpine-rust:/usr/local/cargo "${TEMPDIR}/"
+    podman run --name=official-alpine-rust docker.io/library/rust:alpine sh -c "set -euxo pipefail; apk update && apk add build-base curl; curl --location --silent --show-error --fail https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-$TARGET_ARCH.tgz | gunzip -c | tar -xvvf -; ./cargo-binstall binstall -y cargo-auditable; rm -f \$CARGO_HOME/.crates.toml \$CARGO_HOME/.crates2.json; CARGO_PROFILE_RELEASE_CODEGEN_UNITS=\"1\" CARGO_PROFILE_RELEASE_LTO=\"fat\" OPENSSL_STATIC=1 cargo auditable install $CRATE --version $VERSION"
+    podman cp official-alpine-rust:/usr/local/cargo/.crates2.json "${CARGO_ROOT}/"
+    podman cp official-alpine-rust:/usr/local/cargo/bin "${CARGO_ROOT}/"
+    podman rm -fv official-alpine-rust
 
-    CARGO_BIN_DIR="${TEMPDIR}/cargo/bin"
-    CRATES2_JSON_PATH="${TEMPDIR}/cargo/.crates2.json"
+    CARGO_BIN_DIR="${CARGO_ROOT}/bin"
+    CRATES2_JSON_PATH="${CARGO_ROOT}/.crates2.json"
 elif [ "$TARGET_ARCH" == "aarch64-unknown-linux-gnu" ]; then
     mkdir -p zigfolder
     curl "$(curl -q https://ziglang.org/download/index.json | jq 'to_entries | map([.key, .value])[1][1]["x86_64-linux"] | .tarball' | sed -e 's/^"//' -e 's/"$//')" | tar -xJ -C zigfolder --strip-components 1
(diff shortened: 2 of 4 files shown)


.github/workflows/build-package.yml.template
```diff
@@ -110,15 +110,16 @@ jobs:
         run: |
           (
             cd trigger
-            git tag $CRATE-$VERSION-$TARGET_ARCH
-            git push origin $CRATE-$VERSION-$TARGET_ARCH
+            git tag $CRATE-$VERSION
+            git push origin $CRATE-$VERSION
```
alsuren, line 114: either of these commands might fail now (depending on whether our checkout action also pulls tags+history -  I can't remember)

If they do then just print a message saying this is fine and carry on.

.github/workflows/build-package.yml.template
```diff
@@ -110,15 +110,16 @@ jobs:
...
           )
 
       - name: Releasing assets
-        uses: softprops/action-gh-release@v1
+        uses: svenstaro/upload-release-action@v2
         with:
-          tag_name: $CRATE-$VERSION-$TARGET_ARCH
-          files: |
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          tag_name: $CRATE-$VERSION
```
alsuren, line 121: does this syntax also work for github actions variable substitution? I would expect it to give you a literal `$CRATE-$VERSION`.

alsuren: .... oh yeah: we use sed to do the substituion. Carry on.

build-version.sh
```diff
@@ -25,8 +25,9 @@ if [ "$TARGET_ARCH" == "x86_64-unknown-linux-musl" ]; then
...
+    podman run --name=official-alpine-rust docker.io/library/rust:alpine sh -c "set -euxo pipefail; apk update && apk add build-base curl; curl --location --silent --show-error --fail https://github.com/cargo-bins/cargo-binstall/releases/latest/download/cargo-binstall-$TARGET_ARCH.tgz | gunzip -c | tar -xvvf -; ./cargo-binstall binstall -y cargo-auditable; rm -f \$CARGO_HOME/.crates.toml \$CARGO_HOME/.crates2.json; CARGO_PROFILE_RELEASE_CODEGEN_UNITS=\"1\" CARGO_PROFILE_RELEASE_LTO=\"fat\" OPENSSL_STATIC=1 cargo auditable install $CRATE --version $VERSION"
     podman cp official-alpine-rust:/usr/local/cargo "${TEMPDIR}/"
+    podman rm -fv official-alpine-rust
```
alsuren, line 30: ❤️

.github/workflows/build-package.yml.template
NobodyXu: > does this syntax also work for github actions variable substitution? I would expect it to give you a literal `$CRATE-$VERSION`.

Nope, but this syntax is used here because it is a template.

This file already use this syntax before this PR.

.github/workflows/build-package.yml.template
NobodyXu: Not sure about why this would fail though, since git tag should succeed unless the tag already exists

.github/workflows/build-package.yml.template
```diff
@@ -110,15 +110,16 @@ jobs:
         run: |
```
alsuren, line 110: we *also* need to release in the old location for backwards compatibility. I'd just copy-paste these two blocks below or something. Think about what happens if one succeeds and the other fails though: will we correctly retry?

build-version.sh
```diff
@@ -16,7 +16,7 @@ curl_slowly() {
     sleep 1 && curl --user-agent "cargo-quickinstall build pipeline (alsuren@gmail.com)" "$@"
 }
 
-if [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -I --output /dev/null "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}-${TARGET_ARCH}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
+if [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -I --output /dev/null "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
     echo "${CRATE}/${VERSION}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz already uploaded. Skipping."
     exit 0
 fi
```
alsuren, line 22:
```suggestion
elif [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail --output "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}-${TARGET_ARCH}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
    echo "copy-pasting from the old location rather than building from scratch" >&2
    
    echo "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"
    exit 0
fi
```

... or something

.github/workflows/build-package.yml.template
NobodyXu: > we *also* need to release in the old location for backwards compatibility.

Oh I didn't think of that.
Having backwards compatibility is great, though I do wonder how long will we have to support to old release schema?

There are currently more than 8000 releases on GitHub, I am a bit worries about their growth.

Perhaps we should remove some really old versions, given that it's unlikely anybody wants them and if they really do, they could build these themselves?

> Think about what happens if one succeeds and the other fails though: will we correctly retry?

I changed build-version.sh to check for the new release schema but not the old one, so it would work if we upload the old schema first, then the new schema.

.github/workflows/build-package.yml.template
alsuren: building bat 0.5 for linux will create `bat-0.5` as a tag

build bat 0.5 for macos will try to create `bat-0.5` as a tag but it already exists (or if it doesn't exist locally on the runner due to a sparse checkout, pushing a tag that already exists should fail)

This wasn't a problem with the old code.

NobodyXu: > building bat 0.5 for linux will create `bat-0.5` as a tag
> 
> build bat 0.5 for macos will try to create `bat-0.5` as a tag but it already exists (or if it doesn't exist locally on the runner due to a sparse checkout, pushing a tag that already exists should fail)
> 
> This wasn't a problem with the old code.

I was aware of this, which is why I switch to a different release action, the same one used in cargo-binstall where we upload multiple artifacts from different workflow runs to the same tag.

alsuren: this block of code will fail purely because of the rules of git. This fine, but you need to catch the error

...I've just noticed that we don't have a `set -e` in this block, so it might just carry on if the first command fails (`git tag`, and only bomb out if the second one fails (`git push`). In that case, it may be that this will 'just work". I guess we'll just have to try it and see.

.github/workflows/build-package.yml.template
alsuren: > Having backwards compatibility is great, though I do wonder how long will we have to support to old release schema?

When we release the new quickinstall and binstall clients, we should make them report to the stats server differently or something? This assumes that the stats server is actually working at the moment, which I suspect it might not be.

> There are currently more than 8000 releases on GitHub, I am a bit worries about their growth.

Don't worry about it: rust command-line binaries are pretty small compared to GUI apps and games. If GitHub has a problem with it then they will tell us and we can find a better solution.

alsuren:
```
$ curl 'https://warehouse-clerk-tmp.vercel.app/api/agents?year=2022&month=7'
{"cargo-binstall/0.11.1":4129,"cargo-binstall/0.7.0":4,"cargo-binstall/0.6.2":30,"cargo-binstall/0.9.1":78,"cargo-binstall/0.9.2":2189,"curl/7.79.1":1,"cargo-binstall/0.11.0":76,"cargo-quickinstall client (alsuren@gmail.com)":6652,"cargo-binstall/0.10.0":9250,"cargo-binstall/0.8.0":13}
```
We should probably just put the quickinstall version number in the user agent and that would be enough.

Unfortunately, it looks like 2022/8 is the month that I switched over to collecting stats per-day, and I didn't update https://github.com/alsuren/warehouse-clerk-tmp/blob/master/pages/api/agents.ts .

Also, get-stats.sh is returning {} for me. I wonder if I've hit a size limit on my redis instance or something.

NobodyXu: > > Having backwards compatibility is great, though I do wonder how long will we have to support to old release schema?
> 
> When we release the new quickinstall and binstall clients, we should make them report to the stats server differently or something? This assumes that the stats server is actually working at the moment, which I suspect it might not be.

I tried running the url in get-stats.sh directly on my local machine and it gives me an empty json.

Perhaps there's a token for this?
But it also doesn't give an error.

> > There are currently more than 8000 releases on GitHub, I am a bit worries about their growth.
> 
> Don't worry about it: rust command-line binaries are pretty small compared to GUI apps and games. If GitHub has a problem with it then they will tell us and we can find a better solution.

Thanks, I probably overthink this.

.github/workflows/build-package.yml.template
NobodyXu: Perhaps I should add a set -euxo, it's better to fail ASAP on the first cmd failure

.github/workflows/build-package.yml.template
NobodyXu: > We should probably just put the quickinstall version number in the user agent and that would be enough.

I can do another PR for this.

> Also, get-stats.sh is returning {} for me. I wonder if I've hit a size limit on my redis instance or something.

Yeah I've experienced the same thing.

build-version.sh
```diff
@@ -16,17 +16,23 @@ curl_slowly() {
     sleep 1 && curl --user-agent "cargo-quickinstall build pipeline (alsuren@gmail.com)" "$@"
 }
 
-if [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -I --output /dev/null "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}-${TARGET_ARCH}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
+if [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -I --output /dev/null "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
     echo "${CRATE}/${VERSION}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz already uploaded. Skipping."
     exit 0
+elif [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail --output "${TEMPDIR}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz" "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}-${TARGET_ARCH}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
+    echo "copy-pasting from the old location rather than building from scratch" >&2
+
```
NobodyXu, line 24:
```suggestion

```

.github/workflows/build-package.yml.template
alsuren: I fixed /api/stats and /api/agents. I suspect they've both been broken since july last year and I never noticed. :(

NobodyXu: > I can do another PR for this.

I've created  #166 for this.

.github/workflows/build-package.yml.template
NobodyXu: I've added `set -euxo`.

NobodyXu: Hmm I forgot that `set -euxo` doesn't work on windows.

alsuren: Should work if you put it as the first line of your script block, I think. I've not looked into your build errors yet. Will do in an hour or so when I'm properly awake.

.github/workflows/build-package.yml.template
```diff
@@ -110,15 +110,16 @@ jobs:
...
           (
             cd trigger
-            git tag $CRATE-$VERSION-$TARGET_ARCH
```
alsuren, line 113: I think the only thing that's left is to reinstate the old tag format?

Once that's done, push it to the actions branch a couple of times to test it out and then merge to main and keep an eye on it

NobodyXu: I've reinstalled the old tag format and release format, it should now be backwards compatible.

.github/workflows/build-package.yml.template
NobodyXu: > Should work if you put it as the first line of your script block, I think. I've not looked into your build errors yet. Will do in an hour or so when I'm properly awake.

I've tried [this](https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4333603752/jobs/7566866862) but it fails on windows.

alsuren: You probably want `shell: bash` to make sure windows doesn't use powershell, and then `set -euxo pipefail` inside the script.

NobodyXu: @alsuren Thanks, that works!

.github/workflows/build-package.yml.template
```diff
@@ -109,16 +112,30 @@ jobs:
       - name: Tag release
         run: |
           (
+            set -euxo pipefail
             cd trigger
-            git tag $CRATE-$VERSION-$TARGET_ARCH
-            git push origin $CRATE-$VERSION-$TARGET_ARCH
+            git tag -f $CRATE-$VERSION $CRATE-$VERSION-$TARGET_ARCH
```
alsuren, line 117: we definitely don't want to replace tags that already exist
```suggestion
            git tag $CRATE-$VERSION || echo "$CRATE-$VERSIONtag already exists"
            git tag $CRATE-$VERSION-$TARGET_ARCH
```

NobodyXu: @alsuren What if the `$CRATE-$VERSION-$TARGET_ARCH` already exists but `$CRATE-$VERSION` does not?

.github/workflows/build-package.yml.template
```diff
@@ -109,16 +112,30 @@ jobs:
...
+      - name: Releasing assets in new release format
+        uses: svenstaro/upload-release-action@v2
         with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          tag_name: $CRATE-$VERSION
+          release_name: $CRATE-$VERSION
+          body: build $CRATE@$VERSION
+          file: |
+            ${{github.workspace}}/built.d/$CRATE-$VERSION-$TARGET_ARCH.tar.gz
+          overwrite: true
```
alsuren, line 131: This makes me uncomfortable. Could we put an `if` around this somehow, to skip it if the asset already exists?

.github/workflows/build-package.yml.template
```diff
@@ -109,16 +112,30 @@ jobs:
...
+
+      - name: Releasing assets in old release format
```
alsuren, line 133: Our checking code only checks the new location, so if this step fails we will never retry it. Maybe we should switch the order that we upload things so that this one comes first?

alsuren (approved):
Give it a go on the `actions` branch if you've not done so already, and then fix any issues you find and merge it.

.github/workflows/build-package.yml.template
alsuren:
```suggestion
            git tag $CRATE-$VERSION || echo "$CRATE-$VERSION tag already exists"
            git tag $CRATE-$VERSION-$TARGET_ARCH || echo "$CRATE-$VERSION-$TARGET_ARCH tag already exists"
```

.github/workflows/build-package.yml.template
NobodyXu: I've removed `overwrite` and also updates step "Tag release" to append to `$GITHUB_ENV` if the old or new tag is already present to skip these steps.

.github/workflows/build-package.yml.template
NobodyXu: I've switched the order so that it can be retried on failure.

NobodyXu:
> Give it a go on the `actions` branch if you've not done so already, and then fix any issues you find and merge it.

Uhh how do we trigger a build on `actions` branch again @alsuren ?

NobodyXu:
After reading the workflow weekly, I think I just need to merge this PR into actions to trigger the build.

NobodyXu:
[ build tin-summer 1.21.14 on x86_64-pc-windows-msvc #5889](https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4335879736/jobs/7570770435) failed due to compilation error which is unrelated to the change in this PR:

```
error[E0599]: no method named `mode` found for struct `Permissions` in the current scope
   --> C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\tin-summer-1.21.14\src\walk_parallel\single_threaded.rs:192:39
    |
192 |             if metadata.permissions().mode() == 0o755 || REGEX_GITIGNORE.is_match(path_str) {
    |                                       ^^^^ method not found in `Permissions`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `tin-summer` due to previous error
error: failed to compile `tin-summer v1.21.14`, intermediate artifacts can be found at `C:\Users\RUNNER~1\AppData\Local\Temp\cargo-installwMtL7M`
Error: Process completed with exit code 101.
```

alsuren:
> Uhh how do we trigger a build on actions branch @alsuren ?

just push to origin/actions and it will trigger the cronjob action as a one-off thing each time

NobodyXu:
> > Uhh how do we trigger a build on actions branch @alsuren ?
> 
> just push to origin/actions and it will trigger the cronjob action as a one-off thing each time

Got it, thanks.

Just find another typo in the input to upload-release-action@v2

alsuren:
> P.S. I really want to replace the build-package.yml.template with a regular build-package.yml that is run on workflow_dispatch.

have a go at it on your own fork and see if you can get it to work. We have a lot of stuff hanging off of the "one commit per build attempt" thing, but if you make a PoC we can work out the rest together.

I would really love it if the stats server could also trigger github actions. Using workflow_dispatch might be orthogonal to this though. I've not looked into it properly.

NobodyXu:
> Using workflow_dispatch might be orthogonal to this though. I've not looked into it properly.

Workflow with `workflow_dispatch` can be trigger via GitHub API, so the stat server can definitely trigger it using [this API](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event).

NobodyXu:
@alsuren I've pushed to the `actions` branch again and somehow [the `recheck` workflow](https://github.com/cargo-bins/cargo-quickinstall/actions/runs/4335903347/jobs/7570826128) fails this time:

```
+ git checkout origin/trigger/x86_64-apple-darwin -B trigger/x86_64-apple-darwin
fatal: 'origin/trigger/x86_64-apple-darwin' is not a commit and a branch 'trigger/x86_64-apple-darwin' cannot be created from it
make: *** [Makefile:36: recheck] Error 128
```

There are also a few other errors from `git` in there.

NobodyXu:
Honestly, I have no idea why it would suddenly fail.
Will try replacing that with a regular `git checkout branch`.

NobodyXu:
> Honestly, I have no idea why it would suddenly fail. Will try replacing that with a regular `git checkout branch`.

Tried that, didn't fix the error:

```
+ git checkout trigger/x86_64-apple-darwin
error: pathspec 'trigger/x86_64-apple-darwin' did not match any file(s) known to git
make: *** [Makefile:36: recheck] Error 1
Error: Process completed with exit code 2.
```

NobodyXu:
I think the problem is due to existing tags with duplicate names to the branches:

```
trigger/aarch64-apple-darwin
trigger/aarch64-unknown-linux-gnu
trigger/x86_64-apple-darwin
```

@alsuren Can I delete these tags?

alsuren:
> @alsuren Can I delete these tags?

yes. please delete them. Not sure how they were created.

Issue #164, opened 2023-03-04
switch to hosted influxdb for stats


alsuren:
I think that our redis instance is maybe full or something. I've archived some of the stats onto https://github.com/cargo-bins/cargo-quickinstall/issues/106 but I didn't delete them from redis. Maybe I should do that now.

Crazy plan:

- [ ] delete stats from redis to get us up and running again
- [ ] stop using redis for stats
  - [x] pick influxdb IOx is now the default and the free tier pricing is generous - https://www.influxdata.com/influxdb-cloud-pricing/ (5MB/5 minutes, with 30day retention)
    - [x] set up an org
    - [x] add @NobodyXu (currently sharing the org with my backup mijia-homie monitoring, but I will probably bin that soon)
    - [ ] add other maintainers (email addresses please)
  - [ ] (optional) make the existing stats server proxy http requests onto another stats server that we can all mess about with (or fan out and give everyone their own "staging" server?)
  - [ ] write stats to influxdb using the http api
    - [x] read the docs on influxdb https://docs.influxdata.com/influxdb/cloud-iox/get-started/
    - [x] design schema
      - [x] each install request is an integer count of 1 inserted into a "counts" table, with tags for arch, version, package, client etc.
    - [ ] use http api to do the writing
       - [ ] give vercel the secrets to our influxdb org
       - [ ] write one stat per request (influx wire format is super-simple, but client libraries exist for many languages - we use https://github.com/driftluo/InfluxDBClient-rs in homie-influx but in typescript we could use https://github.com/influxdata/influxdb-client-js if we stick with typescript) 
  - [ ] make github actions cronjob read from influxdb directly
    - [ ] select sum of all counts from the previous day, for the selected arch, grouped by package name
  - [ ] (optional) make a cronjob that archives stats onto a git branch.
  - [ ] (optional) make a thing to roll-up the data into one point per hour per `(crate, version, arch, client)` (summing up the counts for each tag) and set the milisecond-precision database retention policy to 1 week or something

Outstanding questions:

- [ ] Should we switch away from typescript while we are do this?
- [ ] Should we use this as an opportunity to fold the stats server code into this monorepo?

NobodyXu:
@cargo-bins/binstall can you guys all provide your email address please?

passcod:
> make a thing to roll-up the data into one point per hour per (crate, version, arch, client) (summing up the counts for each tag) and set the milisecond-precision database retention policy to 1 week or something

[...]

PR #165, opened 2023-03-04
WIP: Rust-based stats server on fly.io

alsuren:
working towards https://github.com/cargo-bins/cargo-quickinstall/issues/164

- [ ] make the existing stats server proxy http requests onto another stats server that we can all mess about with (or fan out and give everyone their own "staging" server?)
  - [ ] probably urlencode things, like `/record-install?crate=$crate&version=$version&arch=$arch&client=$client`
- [ ] add people to the org
- [ ] dump things into influx
- [ ] make get-stats.sh do a thing


server/fly.io/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM rust:latest as builder
+
+# Make a fake Rust app to keep a cached layer of compiled crates
```
NobodyXu, line 3: Perhaps we can use `cargo-chef` for caching here?

NobodyXu (changes requested):

server/fly.io/Dockerfile
alsuren: Probably no point. The dockerfile in the example repo already does a good enough job of the multi layer build cache dance, so deploying only takes a few seconds.

server/fly.io/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+        .route("/record-install", get(record_install))
+        .route("/record-install", post(record_install));
+
+    // ipv6 + ipv6 any addr
+    let addr = SocketAddr::from(([0, 0, 0, 0, 0, 0, 0, 0], 8080));
+    axum::Server::bind(&addr)
+        .serve(app.into_make_service())
+        .await
+        .unwrap();
+}
```
NobodyXu, line 22: It's better to write it as:

```suggestion
fn main() {
    let rt = tokio::runtime::Runtime::new().unwrap();
    let task = rt.spawn(async move {
        let app = Router::new()
            .route("/", get(root))
            .route("/record-install", get(record_install))
            .route("/record-install", post(record_install));

        // ipv6 + ipv6 any addr
        let addr = SocketAddr::from(([0, 0, 0, 0, 0, 0, 0, 0], 8080));
        axum::Server::bind(&addr)
            .serve(app.into_make_service())
            .await
            .unwrap();
    });
    rt.block_on(task).unwrap();
}
```

so that the main server task can be run on any thread instead of just on `main`, which could be a performance bottleneck.

NobodyXu (changes requested):

server/fly.io/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.0.0"
+authors = ["David Laban <alsuren@gmail.com>"]
+# based on https://github.com/fly-apps/hello-rust by Jerome Gravel-Niquet <jeromegn@gmail.com>
+edition = "2021"
+
+# Make a new workspace so that we get our own Cargo.lock and target dir for the docker build. 
+[workspace]
+
+[dependencies]
+tokio = { version = "1", features = ["full"] }
```
NobodyXu, line 12: We don't need the `full` feature set:

```suggestion
tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
```

We could also remove `macros` dep if you adopt my other suggestion of constructing `tokio::runtime::Runtime` manually to spawn the future.

server/fly.io/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "cargo-quickinstall-stats-server"
+version = "0.0.0"
+authors = ["David Laban <alsuren@gmail.com>"]
+# based on https://github.com/fly-apps/hello-rust by Jerome Gravel-Niquet <jeromegn@gmail.com>
+edition = "2021"
+
+# Make a new workspace so that we get our own Cargo.lock and target dir for the docker build. 
+[workspace]
+
```
NobodyXu, line 10: Using the following profile should reduce binary size and improve performance:

```suggestion

[profile.release]
lto = true
codegen-units = 1
panic = "abort"
strip = "symbols"

```

NobodyXu (changes requested):

server/fly.io/Cargo.toml
NobodyXu: If compilation time is not a concern, then you can also set:

```
lto = true # Use fat lto instead of thin lto, which provides better optimization
codegen-units = 1
```

alsuren: This server is definitely not going to be performance sensitive. The nodejs version was plenty fast enough to handle 10 requests per hour or whatever, and the bottleneck will always be the database.

If anything, adding RUST_BACKTRACE=1 would be more helpful than panic=abort for this kind of thing 

Fly.io free tier doesn't even seem to put things to sleep when idle (unless I'm misreading something)

NobodyXu: Got it, though I still think that setting `lto = "thin"` on release is reasonable.

server/fly.io/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+    // FIXME: make this in main and pass it down or something?
+    let url = get_env("INFLUX_URL");
+    let token = get_env("INFLUX_TOKEN");
+    let org = get_env("INFLUX_ORG");
+    let bucket = get_env("INFLUX_BUCKET");
+    let client = InfluxClient::builder(url, token, org).build().unwrap();
+
+    let mut point = Measurement::builder("counts").field("count", 1);
+    for (tag, value) in &params {
+        point = point.tag(tag, &**value)
```
alsuren, line 46: I had a go at playing with the data:
Screenshot 2023-03-05 at 13 06 31

-- https://eu-central-1-1.aws.cloud2.influxdata.com/orgs/69235d4f38c3e042/notebooks/61ac6f8ba31b7000

Feel free to mess around with it and see what you find.

server/fly.io/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM rust:1-buster as builder
+
+WORKDIR /usr/src/app
+COPY . .
+# Will build and cache the binary and dependent crates in release mode
+RUN --mount=type=cache,target=/usr/local/cargo,from=rust:latest,source=/usr/local/cargo \
+    --mount=type=cache,target=target \
+    cargo build --release && mv ./target/release/cargo-quickinstall-stats-server ./cargo-quickinstall-stats-server
```
NobodyXu, line 8: I think compiling for musl makes more sense?

With musl and everything statically linked, we don't have to use this mounting from rust:latest and can use alpine, or even a scratch as base image

```suggestion
    cargo build --release --target x86_64-unknown-linux-musl && mv ./target/release/cargo-quickinstall-stats-server ./cargo-quickinstall-stats-server
```

server/fly.io/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+[dependencies]
+tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
+axum = "0.6.10"
+influx_db_client = "^0.5.0"
+influxrs = "2.0.1"
+
```
NobodyXu, line 16:
```suggestion
openssl-sys = { version = "0.9", features = ["vendored"] }

```

this will make sure we use statically linked openssl, when compiling for musl.

NobodyXu: For the CVEs, we can just setup dependabot to automatically open PR on CVEs and normal dependencies upgrade.

And we could also have an automatic deploy in GHA.

server/fly.io/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+        axum::Server::bind(&addr)
+            .serve(app.into_make_service())
+            .await
+            .unwrap();
+    });
+    rt.block_on(task).unwrap();
+}
+
+async fn root() -> &'static str {
+    "Hello, World!"
```
NobodyXu, line 30:
```suggestion
    "Hello from quickinstall telemtry! Get /record-install to get the stats, POST on it to report stats!"
```

NobodyXu (changes requested):

server/fly.io/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+
+async fn record_install(Query(params): Query<BTreeMap<String, String>>) -> String {
+    println!("Hi there {params:?}");
+
+    // FIXME: make this in main and pass it down or something?
+    let url = get_env("INFLUX_URL");
+    let token = get_env("INFLUX_TOKEN");
+    let org = get_env("INFLUX_ORG");
+    let bucket = get_env("INFLUX_BUCKET");
+    let client = InfluxClient::builder(url, token, org).build().unwrap();
```
NobodyXu, line 45: With [`std::sync::LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) stablised, we could simply make it global variable:

```suggestion
static INFLUX_CLIENT: LazyLock<InfluxClient> = LazyLock::new(|| {
    let url = get_env("INFLUX_URL");
    let token = get_env("INFLUX_TOKEN");
    let org = get_env("INFLUX_ORG");
    let bucket = get_env("INFLUX_BUCKET");
    InfluxClient::builder(url, token, org).build().unwrap()
});
```

server/fly.io/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+async fn root() -> &'static str {
+    "Hello, World!"
+}
+
+fn get_env(key: &str) -> String {
+    std::env::var(key).expect(&format!("{key} must be set"))
+}
+
+async fn record_install(Query(params): Query<BTreeMap<String, String>>) -> String {
+    println!("Hi there {params:?}");
```
NobodyXu, line 38: I think we could use tracing + + tracing-subscriber + tracing-appender for logging.

tracing is like the crate log, while tracing-appender provides non-blocking writer implementation (to prevent blocking the web server), and tracing-subscriber provides the logging implementation.

tracing-subscriber could also print out json as well to make it machine readable.

server/fly.io/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    let mut point = Measurement::builder("counts").field("count", 1);
+    for (tag, value) in &params {
+        point = point.tag(tag, &**value)
+    }
+    client
+        .write(&bucket, &[point.build().unwrap()])
+        .await
+        .unwrap();
+    format!("Hi there {params:?}")
```
NobodyXu, line 55: I think we can `tokio::spawn` this part, so that the http request would return ASAP (not having to wait for the influx db stuff to be done)

NobodyXu (changes requested):

server/fly.io/Cargo.toml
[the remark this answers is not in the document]
NobodyXu: How about telemtry-collector?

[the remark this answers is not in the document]
alsuren: I like names with fewer syllables. Looks like /stats-server would work as a folder name though.

server/fly.io/Cargo.toml
alsuren: We still need to deal with ca-certificates and other nonsense though (unless that gets vendored as well?). 

Can't we just wait for debian-buster to ship the updated openlls+certs and rebuild the docker image?

(I don't know what the mechanism for that would be though, since we're not exactly doing active development on this server, so it won't just happen as part of day-to-day business)

server/fly.io/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2021"
+
+# Make a new workspace so that we get our own Cargo.lock and target dir for the docker build. 
+[workspace]
+
+[dependencies]
+tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
+axum = "0.6.10"
+influx_db_client = "^0.5.0"
+influxrs = "2.0.1"
```
alsuren, line 15: placeholder to think about https://github.com/ijagberg/influx/issues/4 at some point. This might involve forking their repo temporarily and deploying from the fork

server/fly.io/src/main.rs
alsuren: I think I would prefer human readable logs for this server for now.

I don't think it's ever going to be high volume enough to warrant non-blocking logging: I think that the time for cargo-binstall to start up and make an https connection to the log server will dwarf any time saved by async logging.

I would accept patches to switch to `tracing` though, as long as it doesn't make debugging with the fly.io CLI/web tools harder. I think you still have an outstanding invitation to the fly.io org according to https://fly.io/dashboard/cargo-quickinstall/team. I can try re-sending it if you want.

NobodyXu: > I would accept patches to switch to tracing though, as long as it doesn't make debugging with the fly.io CLI/web tools harder.

👍 I will leave this open so that I don't forget that, though I do think having non-blocking logging is beneficial:

Writing to stdout/stderr is often slower than expected.

> I think you still have an outstanding invitation to the fly.io org according to https://fly.io/dashboard/cargo-quickinstall/team. I can try re-sending it if you want.

Sorry somehow I didn't get it.
Is the invitation linked with my GitHub account, or to my email Jiahao_XU@outlook.com?

Can you send it again please?

server/fly.io/Cargo.toml
alsuren: oh you know what? it might be a problem with how the js server is encoding the url params?

(still not looked into it properly though)

server/fly.io/Cargo.toml
NobodyXu: > We still need to deal with ca-certificates and other nonsense though (unless that gets vendored as well?). 

Maybe we can try https://github.com/GoogleContainerTools/distroless

> Distroless images are very small. The smallest distroless image, gcr.io/distroless/static-debian12, is around 2 MiB. That's about 50% of the size of alpine (~5 MiB), and less than 2% of the size of debian (124 MiB).

> Can't we just wait for debian-buster to ship the updated openlls+certs and rebuild the docker image?

We won't get any notification for the image, but we can have dependabot automatically opening PR to update rust dependencies and trigger redeployment.

server/fly.io/src/main.rs
NobodyXu: Oops I was a bit silly, thought get is for getting statistics but no it's not.

server/fly.io/src/main.rs
alsuren: I've had trouble recently with things like google cloud run suspending your compute as soon as you return from your http request, so I'm a bit wary of doing this (yes I know the vercel stats server does all sorts of nonsense like this, and fly.io isn't as aggressive as cloud run). I think I prefer structured concurrency these days.

server/fly.io/Dockerfile
alsuren: Hrm. After fighting with debian for a bit, I might be coming round to your way of thinking. I'm going to plug away at it after lunch for a bit and see where I get to.

server/fly.io/src/main.rs
alsuren: The original request was sent to Jiahao_XU@outlook.com . I revoked it and sent it again. Maybe it went to spam?

NobodyXu: Thanks, I've received it now.

I looked it up in my spam but somehow didn't find it (maybe missed it?

stats-server/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "cargo-quickinstall-stats-server"
+version = "0.0.0"
+authors = ["David Laban <alsuren@gmail.com>"]
+# based on https://github.com/fly-apps/hello-rust by Jerome Gravel-Niquet <jeromegn@gmail.com>
+edition = "2021"
+
+# Make a new workspace so that we get our own Cargo.lock and target dir for the docker build. 
+[workspace]
```
alsuren, line 9: For some reason my vscode is no longer giving any hints for anything in this folder unless I explicitly open only this folder in its own instance. I'm sure it worked before, and for other projects, but I can't remember what I did last time.

Probably I will merge this as is, but if this causes problems for anyone else then tell me and I'll try to address it properly.

server/fly.io/Dockerfile
alsuren: turns out I got it working with debian again. I'm going to backlog the problem of reacting to CVEs, I think.

NobodyXu: Well, if you use vendored openssl, then whenever a new openssl is published, a new openssl-src will be published, and dependabot is able to create a PR to update them.

With auto deployment in GHA, it'd be a smooth process.

server/fly.io/Cargo.toml
alsuren: I'm going to punt the CVE thing for now (https://github.com/cargo-bins/cargo-quickinstall/issues/277)

re container size: we're still much smaller than anything nodejs or python based, so I'm going to defer this as well. Fly.io hasn't exactly complained, and I've not looked into our request latency numbers, so I don't know if cold starts are likely to be a problem (I assume they wait for us to pass health checks before dumping us in the load balancer, so I doubt it will be a problem).

server/fly.io/src/main.rs
alsuren: backlogged logging as https://github.com/cargo-bins/cargo-quickinstall/issues/278

server/fly.io/src/main.rs
alsuren: backlogged shared grafana dashboards as https://github.com/cargo-bins/cargo-quickinstall/issues/279

server/fly.io/Dockerfile
alsuren: backlogged CVE stuff as (https://github.com/cargo-bins/cargo-quickinstall/issues/277) for now

server/fly.io/Cargo.toml
alsuren: Fixed in https://github.com/ijagberg/influx/issues/5 . Depending on my fork for now.

.github/workflows/deploy-stats-server.yml
alsuren, line 1: Roughly following https://fly.io/docs/launch/continuous-deployment-with-github-actions/ but using an environment to hold the secret, so you can only access the fly secret if you can push to main.

This was tested on a previous commit.

Logs exist here: https://github.com/cargo-bins/cargo-quickinstall/actions/runs/10761262315/job/29840247483?pr=165

NobodyXu (changes requested):

NobodyXu (approved):

PR #169, opened 2023-03-06
Add support of new release schema to `cargo-quickinstall`

NobodyXu:


alsuren (approved):
Looks good.

Let's wait until all of the popular packages have been reuploaded in the new scheme and then rip out the fallback code. Ideally we would do this before the next release, so we never ship the fallback code.

You can speed things up by checking out main and calling ./trigger-package-build.sh in a sleep loop (we could probably add a way to make it stop when it loops back to building cargo-quickinstall, but if it loops a couple of times that's probably fine)

PR #174, opened 2023-03-06
Refactor computation of `POPULAR_CRATES` in `next-unbuilt-package.sh`

NobodyXu:

### DIFF
--- a/dedup-and-exclude.py
+++ b/dedup-and-exclude.py
@@ -1,0 +1,36 @@
+#!/usr/bin/env python3
+
+import sys
+
+def strip_whitespace(iterable):
+    for s in iterable:
+        s = s.strip()
+        if len(s) != 0:
+            yield s
+
+def stderr(s):
+    print(s, file=sys.stderr, flush=True)
+
+if len(sys.argv) == 3:
+    with open(sys.argv[1]) as f:
+        s = set(strip_whitespace(f))
+    max_lines = int(sys.argv[2])
+else:
+    program = sys.argv[0]
+    stderr(f"Usage: {program}: /path/to/excludes max-num-of-lines-to-output")
+    sys.exit(1)
+
+dup = set()
+cnt = 0
+
+for line in sys.stdin:
+    line = line.strip()
+
+    if line in s:
+        stderr(f"skipping {line} because it has failed too many times")
+    elif len(line) != 0 and line not in dup:
+        print(line)
+        dup.add(line)
+        cnt += 1
+        if cnt == max_lines:
+            break
--- a/dedup-and-exclude.py
+++ b/dedup-and-exclude.py
@@ -13,24 +13,37 @@ def stderr(s):
 
 if len(sys.argv) == 3:
     with open(sys.argv[1]) as f:
-        s = set(strip_whitespace(f))
+        exclude = set(strip_whitespace(f))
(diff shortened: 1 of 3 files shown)


NobodyXu:
I can't understand why `b3sum` isn't rebuilt on CI.
Is it in excludes.txt or something?
Also, is the `popular.txt` in the `trigger/*` branch too old?

get-stats.sh
```diff
@@ -4,19 +4,9 @@ set -euo pipefail
...
     --silent \
     --show-error \
     -XGET \
-    "https://warehouse-clerk-tmp.vercel.app/api/stats" | (
-    # Slight hack: if TARGET_ARCH is specified then just print crate names one per line.
-    # Otherwise print all counts as json.
-    if [[ "${TARGET_ARCH:-}" != "" ]]; then
-        jq -r 'keys | .[]' |
-            (grep -F "${TARGET_ARCH:-}" || true) |
-            sed 's:/.*::' || (echo "Invalid json from stats server. Probably rate limit error. Skipping for now." >&2)
```
alsuren, line 18: these `|| true` and `|| echo` things are there because the stats server might have hit its rate limit or the grep might return 0 lines (in which case it "fails" and -o pipefail causes that failure to propagate).

NobodyXu: Done!

NobodyXu:
Thanks for explanation @alsuren

dedup-and-exclude.py
```diff
@@ -0,0 +1,36 @@
...
+def strip_whitespace(iterable):
+    for s in iterable:
+        s = s.strip()
+        if len(s) != 0:
+            yield s
+
+def stderr(s):
+    print(s, file=sys.stderr, flush=True)
+
+if len(sys.argv) == 3:
```
alsuren, line 14:
```suggestion
def main(arg1, arg2):
...

if __name__ == "__main__":
    main(sys.argv[0], ...)
```

dedup-and-exclude.py
```diff
@@ -0,0 +1,36 @@
...
+    stderr(f"Usage: {program}: /path/to/excludes max-num-of-lines-to-output")
+    sys.exit(1)
+
+dup = set()
+cnt = 0
+
+for line in sys.stdin:
+    line = line.strip()
+
+    if line in s:
```
alsuren, line 29: first loop s is consumed. It is empty on second loop.

dedup-and-exclude.py
```diff
@@ -0,0 +1,36 @@
...
+def stderr(s):
+    print(s, file=sys.stderr, flush=True)
+
+if len(sys.argv) == 3:
+    with open(sys.argv[1]) as f:
+        s = set(strip_whitespace(f))
+    max_lines = int(sys.argv[2])
+else:
+    program = sys.argv[0]
+    stderr(f"Usage: {program}: /path/to/excludes max-num-of-lines-to-output")
```
alsuren, line 20: (optional) inline this because it is only used once

alsuren (approved):

dedup-and-exclude.py
NobodyXu: The function `stderr` is actually used in the loop as well.

NobodyXu: The function `stderr` is actually used in the loop as well.

PR #175, opened 2023-03-06
Add new workflow build-package with workflow_dispatch

NobodyXu:
This replaces the old workflow that is triggered on pushing to `trigger/$TARGET_ARCH`.

The workflow can be triggered on github webpage, using `gh` cli, or using GitHub Restful API through `curl` or `reqwest` in rust.

I've verified that it is working and can trigger the build as expected.

TODO:
 - [x] concurrency-group, using `${{ inputs.target_arch }}` according to [here](https://stackoverflow.com/questions/74242156/github-actions-concurrency)
 - [x] create excludes.txt on the fly using `prints-build-excludes.sh`, excluding any crate that fails to build for 5 times or more.
 - [x] [shuf](https://unix.stackexchange.com/questions/167013/random-permutation-of-lines-of-text) the `POPULAR_CRATES`.


next-unbuilt-package.sh
```diff
@@ -61,14 +61,11 @@ for CRATE in $POPULAR_CRATES; do
         echo "${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz already uploaded. Keep going." 1>&2
```
alsuren, line 61: Higher up in this file: where does START_AFTER_CRATE come from after this change? (it currently comes from the last commit on trigger/*)

trigger-package-build.sh
```diff
@@ -82,74 +75,28 @@ main() {
             git add "$EXCLUDE_FILE"
```
alsuren, line 75: print-build-excludes.sh also depends on the git history. You will need to use the GitHub actions query API for this or something.

next-unbuilt-package.sh
NobodyXu: I removed `START_AFTER_CRATE` because I'm not sure what is the use of it.
It seems to me it is easier just removing it.

trigger-package-build.sh
NobodyXu: Got it, I would rewrite `print-build-excludes.sh`.

NobodyXu: It seems that `gh run` does not give me any information about the `inputs` provided to the workflow.

I'm now thinking of directly creates a commit to the excludes.txt on failure of `bulid-package`.

NobodyXu: I've implemented a new mechanism for generating excludes, however since none of the build has failed yet, it's not tested.

next-unbuilt-package.sh
NobodyXu: I think with #174 , we should be good without `START_AFTER_CRATE`.

alsuren: I will try to describe the problems that START_AFTER_CRATE solves shortly. I will be making diagrams on https://app.excalidraw.com/l/4NJeenUSxBH/48aT0OfmkFU if you want to follow along. I will paste them in here when I'm done.

next-unbuilt-package.sh
```diff
@@ -57,18 +59,15 @@ for CRATE in $POPULAR_CRATES; do
...
-    if curl_slowly --location --fail -I --output /dev/null "https://github.com/cargo-bins/cargo-quickinstall/releases/download/${CRATE}-${VERSION}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
+    if curl_slowly --location --fail -I --output /dev/null "${REPO}/releases/download/${CRATE}-${VERSION}/${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz"; then
         echo "${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz already uploaded. Keep going." 1>&2
     else
         echo "${CRATE}-${VERSION}-${TARGET_ARCH}.tar.gz needs building" 1>&2
-        echo "::set-output name=crate_to_build::$CRATE"
-        echo "::set-output name=version_to_build::$VERSION"
-        echo "::set-output name=arch_to_build::$TARGET_ARCH"
+        echo "{\"crate\": \"$CRATE\", \"version\": \"$VERSION\", \"target_arch\": \"$TARGET_ARCH\"}"
         exit 0
```
NobodyXu, line 67:
```suggestion
```

next-unbuilt-package.sh
alsuren: for posterity:

START_AFTER_CRATE-problem-1 excalidraw
START_AFTER_CRATE-problem-2 excalidraw
START_AFTER_CRATE-solution-1 excalidraw

NobodyXu:
image

image

Seems to be working, but since I just adjusted the `concurrency.group` keys again, will have to wait and observe the new build-package.yml workflows triggered

NobodyXu:
Confirmed that the `concurrency.group` works, but with a catch:

```
Canceling since a higher priority waiting request for 'build-package-x86_64-pc-windows-msvc' exists
```

it could be cancelled if there is a higher priority waiting requests.
I would have to investigate on this and disable this cancellation.

NobodyXu:
> Confirmed that the `concurrency.group` works, but with a catch:
> 
> ```
> Canceling since a higher priority waiting request for 'build-package-x86_64-pc-windows-msvc' exists
> ```
> 
> it could be cancelled if there is a higher priority waiting requests. I would have to investigate on this and disable this cancellation.

This problem is discussed [here](https://github.com/community/community/discussions/5435).
I would instead change the `concurrency.group` to contain the crate, the version and the target_arch to avoid this cancellation.

NobodyXu:
> > Confirmed that the `concurrency.group` works, but with a catch:
> > ```
> > Canceling since a higher priority waiting request for 'build-package-x86_64-pc-windows-msvc' exists
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it could be cancelled if there is a higher priority waiting requests. I would have to investigate on this and disable this cancellation.
> 
> This problem is discussed [here](https://github.com/community/community/discussions/5435). I would instead change the `concurrency.group` to contain the crate, the version and the target_arch to avoid this cancellation.

While this is not as good as we originally discussed to avoid overloading GHA, this still prevents [case 1](https://user-images.githubusercontent.com/254647/223373907-25dca6c8-bd84-49a4-b265-be408187c55c.png).

alsuren (approved):
I've not had a chance to look at this again since we last spoke, but I trust that you understand the failure modes, so take another look over your own code and then ship it and keep an eye on it.

If anything goes wrong then hit the revert button on this PR and it should hopefully all calm down.

Good luck.

PR #202, opened 2023-03-14
Release cargo-quickinstall v0.2.9

NobodyXu:


NobodyXu:
@alsuren I think it's time for a v0.2.9 release.

I decided that we'd better keep the fallback to old release schema for now, since our CI only builds the `max_stable_version` of the crates and does not build the old versions of the crates.

alsuren (approved):
Sounds good.

PR #209, opened 2023-04-16
Fix building gitoxide

NobodyXu:
`gitoxide` is crate intended to be an alternative of `git`.
It currently fails on cross compilation due to lack of system-wide openssl installation.

It's quite popular, so I think we should fix building `gitoxide` in quickinstall.

According to @Byron in https://github.com/Byron/gitoxide/issues/809 , we can fix the by building gitoxide with `--no-default-features --featulres max-pure`.

 - Add new input `no_default_features` to `build-package.yml`
 - Build gitoxide with `--no-default-features --features max-pure`
 - Add `\.DS_Store` to `.gitignore`

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -27,6 +27,10 @@ on:
         description: Features to enable
         default: ""
         type: string
+      no_default_features:
+        description: Disable default features
+        default: ""
+        type: string
 
 concurrency:
   group: build-package-${{ github.event.inputs.crate }}-${{ github.event.inputs.version }}-${{ github.event.inputs.target_arch }}
@@ -36,6 +40,7 @@ env:
   TARGET_ARCH: ${{ inputs.target_arch }}
   VERSION: ${{ inputs.version }}
   FEATURES: ${{ inputs.features }}
+  NO_DEFAULT_FEATURES: ${{ inputs.NO_DEFAULT_FEATURES }}
   BUILD_DIR: ${{ github.workspace }}/built.d
 
 jobs:
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .env
 .idea
 .vagrant
+\.DS_Store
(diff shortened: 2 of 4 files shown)


Byron:
Thanks for making `gitoxide` installations work with `cargo quickinstall`, appreciated!

> `gitoxide` is crate intended to be an alternative of `git`.

Let me clarify that the command-line tools, `gix` and `ein`, that come with the `gitoxide` crate both have their own purpose. `gix` might seem like an alternative to `git` even though it's intended as development tool which happens to be able to do a few things better or faster than `git` can. It will, however, be lacking a lot of features users of `git` will want or need for their daily workflow.

`ein` may one day be able to drive developer interactions with `git` repositories all on its own, but that work is in the far future, if it's ever realised at all.

Thus I think the term 'alternative' might give rise to expectations that will remain unfulfilled. That said, however, both tools have their uses and with `cargo quick-install` it's easier to see if they fit a niche for you 🎉 (*thanks again*).

NobodyXu:
Thanks for the explanation!

I read on somewhere else (maybe reddit) that gitoxide is not intended to be a full replacement for git, however I don't know what else should I said in the description other than gitoxide being an alternative of git, so I just wrote that in the description.

I probably should said that `gitoxide` is an alternative libgit/git implementation, which should be more clear than the original.

> That said, however, both tools have their uses and with cargo quick-install it's easier to see if they fit a niche for you 🎉 (thanks again).

You can also use `gitoxide` using [`cargo-binstall`], a project that is designed to download pre-built artifacts from upstream (the dev, e.g. GitHub Release), fallback to quickinstall if not found and finally fallback to `cargo-install`.

It is intended to be a replacement for `cargo-install` and actually updates manifest `$CARGO_HOME/.crates.toml`, in the same way as `cargo-install`.

I encourage you to try `cargo-binstall`, it can be used through `cargo install-updates -a` once you installed it on your system to speedup updating binary crates you installed.

Feel free to ask me any question or report any new feature request/bug in [`cargo-binstall`]!

[`cargo-binstall`]: https://github.com/cargo-bins/cargo-binstall

PR #230, opened 2023-09-10
feat: Add binstall signing support

NobodyXu:


NobodyXu:
@passcod Need some help here:

I've added `MINISIGN_PRIVATE_KEY` secret key yet I can't access it in CI https://github.com/cargo-bins/cargo-quickinstall/actions/runs/6137161396/job/16653023682?pr=230#step:9:14

Issue #251, opened 2024-06-14
Sanbox build pipeline by using docker


NobodyXu:
While docker isn't exactly safe, it is a good start.

We could then use fieecracker/gVisor to further sandbox the build.

Lastly, we could use RUSTC_WRAPPER to run each rustc invocation in a separate container.

PR #252, opened 2024-06-17
Sandbox linux builds using docker and gVisor

NobodyXu:
Related: #251


NobodyXu:
cc @alsuren @Milo123459 opened this PR to sandbox builds on linux using docker and gVisor.

FYI gVisor is a container runtime created by Google, that minimizes the attack surface by implementing minimal amounts of syscalls in userspace, and has its own userspace overlayfs and network stack.

On GHA it uses seccomp's systrap to trap syscalls and emulates them, on bare metal it can use KVM https://gvisor.dev/docs/architecture_guide/platforms/.

alsuren:
(I've not read the code yet)

What is the threat model that this is mitigating against? 

Currently we have a sandbox around the job that builds the crate (each job in GitHub actions runs in an entirely separate VM): it can only communicate with the rest of the world by putting things in the job's build artifacts, and then we have a separate job (VM) that fetches the artifact and pushes it to GitHub releases.

As soon as we run our first untrusted code, we treat the whole job runner as compromised, so I don't think we gain anything by adding another layer of sandboxing inside that.

NobodyXu:
> What is the threat model that this is mitigating against?

I think each job in GitHub actions still have access to some environment variables (implicitly defined) and some github context, plus root access to the VM.

For example, the cargo-build can access `GITHUB_ENV`, `GITHUB_OUTPUT`.

cc @alsuren For the upload action, my understanding is that the action, written in JS, will be run on the VM itself.

Otherwise any action written in JS won't have access to the filesystem in the VM and won't be able to upload anything, and AFAIK action written in JS just `import fs` as normal, so I doubt there's any remote FS access there.

Since the action needs to upload data, it must have access to the `secrets.GITHUB_TOKEN`, which is a bit dangerous:

At the very least, having access to it allow attackers to launch a DDOS attack using the GitHub token, prevent any job from running by exhausting the points of the restful/graphql API.

The docker container:
 - mounts the image (cached in GHA) as read-only
 - prevent it from gaining any new privileges
 - run it inside gVisor with a userspace FS and network stack
 - controls what environment variables can be accessed (without docker, you could just use `/proc` to access them anyway).

In the future would could also limit network access.

Another thing I planned to do is, run each `rustc` instance in their own, untouchable container sandboxed by gVisor.

In that way, it is then possible to use sccache to cache them.

It would also provide sandbox for proc-macros, though unfortunately I'm still unable to figure out how to sandbox build-script.

.github/workflows/build-package.yml
```diff
@@ -87,94 +87,84 @@ env:
...
+            echo "NEEDS_BUILD=true" >> "$GITHUB_OUTPUT"
+          fi
+          
   build-popular-package:
-    name: build-popular-package-${{ inputs.crate }}-${{ inputs.version }}-${{ inputs.target_arch }}-${{ inputs.features }}
+    needs: check-if-already-built
+    if: inputs.always_build != '' || needs.check-if-already-built.outputs.NEEDS_BUILD == 'true'
+    name: Build package ${{ inputs.crate }}-${{ inputs.version }}-${{ inputs.target_arch }}-${{ inputs.features }}
     runs-on: ${{ inputs.build_os }}
     permissions: {}
```
alsuren, line 113: (splitting my reply into line comments so that we can have multiple threads, because there's a lot going on)

> Otherwise any action written in JS won't have access to the filesystem in the VM and won't be able to upload anything, and AFAIK action written in JS just import fs as normal, so I doubt there's any remote FS access there.
> 
> Since the action needs to upload data, it must have access to the secrets.GITHUB_TOKEN, which is a bit dangerous:
>
> At the very least, having access to it allow attackers to launch a DDOS attack using the GitHub token, prevent any job from running by exhausting the points of the restful/graphql API.

My understanding was that it should not have enough scopes to do anything, after I set this option. You are right that the actions/upload-artifact action clearly manages to use **some** kind of github api to do its uploading. It is worth reading into how that works, and seeing if we can replicate it using something like:

```yaml
- name: Setup tmate session
  uses: mxschmitt/action-tmate@v3
  with:
    limit-access-to-actor: true
```

to drop us into a tmux session on the runners, so we can fuck around in bash.

alsuren: if it is possible to do anything non-job-local (e.g. writing to the @actions/cache cache, or burning our repo-wide github api rate limits) from a tmate session in a job with `permissions: {}` then we might need to pause the builders for a bit until we can do a proper audit of how we're doing things

.github/workflows/actions/build-version/action.yml
```diff
@@ -0,0 +1,135 @@
...
+      uses: docker/setup-buildx-action@v3
+
+    - name: Build docker image for package building
+      if: runner.os == 'Linux'
+      uses: docker/build-push-action@v5
+      with:
+        context: .
+        load: true
+        tags: package-builder:latest
+        cache-from: type=gha
```
alsuren, line 96: it's a bit scary that we can write to a cache here and potentially have it read by another package's build

.github/workflows/actions/build-version/action.yml
```diff
@@ -0,0 +1,135 @@
...
+          no_default_features='--no-default-features'
+        fi
+        echo "CARGO_ARGS=$no_default_features $feature_flag $features" >> "$GITHUB_ENV"
+        echo "TEMPDIR=$(mktemp -d)" >> "$GITHUB_ENV"
+        echo CARGO_HOME="$(mktemp -d 2>/dev/null || mktemp -d -t 'cargo-root')" >> "$GITHUB_ENV"
+      shell: bash
+
+    - name: Build packages
+      if: runner.os != 'Linux'
+      run: ./build-version.sh
```
alsuren, line 78: > > What is the threat model that this is mitigating against?
>
> I think each job in GitHub actions still have access to some environment variables (implicitly defined) and some github context

If getting access to these things is game-over then we should turn off all non-linux builders and cross-compile everything from linux or something. I really hope it's not though (read https://alsuren.github.io/2022/07/10/cargo-quickinstall.html#github-actions-gymnastics for some history about why I'm treating it as basically fine)

.github/workflows/actions/build-version/action.yml
NobodyXu: This is for building `Dockerfile` written by me, it only downloads the dependencies and never do any `cargo build`

.github/workflows/build-package.yml
NobodyXu: https://github.com/actions/toolkit/blob/361a115e538ac6d8eb06cc47f3fcecce557d04c8/packages/artifact/src/internal/shared/config.ts#L10

It seems that the github token is passed to the JS action as environment variable

alsuren: Okay, so the next question is how much can that token actually do? 

The artifact storage seems to be sandboxed per run or something, and actions/cache is per branch, with the ability to read from your base branch's cache but not write to it (unless you set something in your permissions, here). In our case I think all builds on the same arch happen from the same branch, so I guess we should treat anything we read from the cache as untrusted? I had previously assumed that having permissions: {} meant that you couldn't write to the cache at all, so we might be safe there, as long as we don't merge your docker cache stuff (see other thread)

.github/workflows/actions/build-version/action.yml
alsuren: The artifact storage seems to be sandboxed per run or something, and actions/cache is per branch, with the ability to read from your base branch's cache but not write to it (unless you set something in your permissions, here). In our case I think all builds on the same arch happen from the same branch, so I guess we should treat anything we read from the cache as untrusted? 

I don't think it matters who wrote the code that is writing to the cache: if the cache is writeable from an action that's happening on this worker (for this branch) then we can't trust anything that's returned from the cache.

.github/workflows/build-package.yml
NobodyXu: They can definitely keep issuing new restful/graphql API until it reaches rate limit, in that case, all our jobs would be unable to run due to exhaustion of rate limit.

`secrets.GITHUB_TOKEN` has a primary rate limit of 1000 per hour, the secondary rate limit of 100 concurrent API requests, once it is exhaustes all other jobs could stop working.

https://docs.github.com/en/graphql/overview/rate-limits-and-node-limits-for-the-graphql-api#secondary-rate-limits

.github/workflows/actions/build-version/action.yml
NobodyXu: > I don't think it matters who wrote the code that is writing to the cache: if the cache is writeable from an action that's happening on this worker (for this branch) then we can't trust anything that's returned from the cache.

That's a good point, while sandboxing should prevent them from accessing any network or the token, there's always a possibility that it will be broken.

I will move it into a seperate job and then export the created image as artefacts.

NobodyXu: I've moved it to a separate job, now it should be ok.

.github/workflows/build-package.yml
alsuren: Right, but this is only for the duration of the build, so it's only going to affect the packages that happen to be building for other architectures at the same time, until this job times out (1h?).

If you want to DoS the quickinstall workers, you could just publish 1000 empty packages on crates.io and then request them all from the stats server.

I don't think that the consequences of this attack justify the complexity of the mitigation, especially if the mitigation does not affect windows and macos.

Don't get me wrong: I am learning a lot from this PR, and it's interesting to explore the possibilities for sandboxed builds. I am just not convinced that the threat model has been thought through enough to validate that this mitigation is necessary+sufficient.

I'm happy to jump on a call this weekend and chat things through (send me a time in Europe/London timezone and I'll set up a call). I've also just handed in my notice to quit at work, so I might have some time off between jobs in a few weeks, for daytime pairing.

.github/workflows/actions/build-version/action.yml
alsuren: @NobodyXu let's make sure we've addressed this thread too.

.github/workflows/build-package.yml
NobodyXu: I'm a bit busy at weekends and in fact, have more time during Mondays to Thursdays to discuss about this.
I'm in Sydney, so there's about 9 hours of difference.

.github/workflows/actions/build-version/action.yml
NobodyXu: I think this is related to what we discussed in https://github.com/cargo-bins/cargo-quickinstall/pull/252#discussion_r1645007351

Regarding cross compilation, I think zig-cc could help us achieve cross compiling to macos and mingw-gnu on Linux.

NobodyXu:
The failure is because the old version of zig write to its our installation directory, which is immutable.

The blocker for newer zig is that it rejects unknown linker flag `--undefined` which is used by `cargo-auditable`.

I think it makes sense to remove `cargo-auditable` for now, so that we can upgrade to latest release, since the build-process could modify whatever it likes and leave process behind, it is trivial to modify it to something else to prevemt auditing.

NobodyXu:
cc @alsuren I just discovered https://github.com/cackle-rs/cackle

A tool with sandboxing and code auditing built-in, which seems better than my homebrew solution.

I will close this PR, and open a new PR to refactor/rework the build-system of quickinstall.

Issue #268, opened 2024-08-20
402 Payment Required error


paul-hansen:
Getting this error just now in our CI
```
Run cargo binstall -y --force cargo-leptos
 INFO resolve: Resolving package: 'cargo-leptos'
 WARN Failed to send quickinstall report for package cargo-leptos-0.2.19-x86_64-unknown-linux-gnu: Failed to download from remote: could not HEAD https://warehouse-clerk-tmp.vercel.app/api/crate/cargo-leptos-0.2.19-x86_64-unknown-linux-gnu.tar.gz: HTTP status client error (402 Payment Required) for url (https://warehouse-clerk-tmp.vercel.app/api/crate/cargo-leptos-0.2.19-x86_64-unknown-linux-gnu.tar.gz)
 WARN Failed to send quickinstall report for package cargo-leptos-0.2.19-x86_64-unknown-linux-musl: Failed to download from remote: could not HEAD https://warehouse-clerk-tmp.vercel.app/api/crate/cargo-leptos-0.2.19-x86_64-unknown-linux-musl.tar.gz: HTTP status client error (402 Payment Required) for url (https://warehouse-clerk-tmp.vercel.app/api/crate/cargo-leptos-0.2.19-x86_64-unknown-linux-musl.tar.gz)
```

It then builds it from source. It worked earlier today without this error.

alsuren:
Yeah, vercel said we hit our http handler time limit. We could probably talk to vercel and get that bumped.

NobodyXu:
Thank you!

That's good to hear!

paul-hansen:
Seems to be working again, closing. Thanks!

Feel free to reopen if you want to use this issue as a reminder to ask them to bump the limit if you haven't yet or anything.

s373r:
Also ran into this problem. A clipping of logs that you may find useful:
```
...
+ ./cargo-binstall -y --force cargo-binstall
 INFO resolve: Resolving package: 'cargo-binstall'
 WARN Failed to send quickinstall report for package cargo-binstall-1.10.3-x86_64-apple-darwin: Failed to download from remote: could not HEAD https://warehouse-clerk-tmp.vercel.app/api/crate/cargo-binstall-1.10.3-x86_64-apple-darwin.tar.gz: HTTP status client error (402 Payment Required) for url (https://warehouse-clerk-tmp.vercel.app/api/crate/cargo-binstall-1.10.3-x86_64-apple-darwin.tar.gz)
[...]

PR #275, opened 2024-09-06
rewrite Cronjob build triggering logic in python; read from influxdb

alsuren:
- [x] rewrite a bunch of the bash build scripts in python
  - [x] dependabot (completely untested)
  - [x] automatic installation of deps when you run relevant `make` rules
  - [x] delete old bash scripts
- [x] switch to reading from influxdb cloud directly, instead of redis via the webserver. See https://github.com/cargo-bins/cargo-quickinstall/pull/165
  - [ ] rebase https://github.com/cargo-bins/cargo-quickinstall/pull/165 and address comments
  - [ ] you now need INFLUXDB_TOKEN to read these stats.
    - [ ] I have added one to our github secrets (with read-only creds)
    - [ ] I need to put everyone in the influx org (different one from the one I already invited everyone too: sorry)
    - [x] I can either make a new public read-only token and put it in the repo (can revoke it if someone eats our rate limits or something) or add a /stats endpoint on the server like the old one had? - you can use `p1YOOpyPWwrcJ5wqeBqpsIVPGMiKDHg_PUkUxFdwl9lW6W5jHLdxL67Fsz4STm1yDd8pwH42YHuX7ZxjRkOu4w==
` for now. I will think about revoking it later.
- [ ] not sure where my print() statements output is disappearing to in the github actions output :(


README.md
```diff
@@ -45,7 +45,7 @@ Edit the command however you need, and paste it into your CI pipeline.
 
 ## Supported targets
 
-Check [supported-targets](/supported-targets) for lists of targets quickinstall
+Check [supported-targets](/cronjob_scripts/trigger-package-build.py) for lists of targets quickinstall
```
NobodyXu, line 48: Can we keep the `supported-targets`?

cargo-binstall uses it to avoid hitting github APIs, if binstall knows that quickinstall does not support the target.

NobodyXu (changes requested):

Makefile
```diff
@@ -2,38 +2,49 @@
 publish: release ## alias for `make release`
 
 .PHONY: release
-release: ## Publish a new release
+release: cronjob_scripts/.python-deps-updated.timestamp ## Publish a new release
 	(cd cargo-quickinstall/ && cargo release patch --execute --no-push)
 	git push origin HEAD:release --tags
 	make recheck
```
NobodyXu, line 8: I think `$(MAKE) recheck` is better/more canonical way of invoking another makefile?

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,217 @@
...
+    if target_arch in TARGET_ARCH_TO_BUILD_OS:
+        return [target_arch]
+
+    if target_arch == "all":
+        return list(TARGET_ARCH_TO_BUILD_OS.keys())
+
+    try:
+        rustc_version_output = subprocess.run(
+            ["rustc", "--version", "--verbose"], capture_output=True, text=True
+        )
```
NobodyXu, line 136:
```suggestion
    rustc_version_output = subprocess.run(
         
["rustc", "--version", "--verbose"], capture_output=True, text=True
    )

    try:
```

Since the `except` access `rustc_version_output`, it can't handle the error from `subprocess.run`

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,217 @@
...
+    except:
+        print(f"Failed to get latest version for {crate}")
+        return None
+
+    url = f"{repo_url}/releases/download/{crate}-{version['vers']}/{crate}-{version['vers']}-{target_arch}.tar.gz"
+    response = requests.head(url, allow_redirects=True)
+    if response.ok:
+        print(f"{url} already uploaded")
+        return None
+    # FIXME: handle rate limit errors?
```
NobodyXu, line 211: Given that we already get all tags from repo, I suppose we can use `git describe --match *`?

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+    failures = Counter()
+
+    failure_filenames = list(Path(tracking_worktree_path).glob("*-*-*"))
+    failure_filenames.sort(reverse=True)
+
+    for failure_filename in failure_filenames[:days]:
+        with open(failure_filename, "r") as file:
+            failures.update(
+                Counter(line.strip() for line in file if line.strip() != "")
```
NobodyXu, line 170:
```suggestion
                Counter(stripped for line in file if (stripped := line.strip()))
```

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,217 @@
...
+    check_limit = int(os.environ.get("CRATE_CHECK_LIMIT", 5))
+    # Note that the old next-unbuilt-package.sh script actually checked the top 4 * CRATE_CHECK_LIMIT
+    # crates from get-stats.sh + popular-crates.txt rather than shuffling the whole lot before checking.
+    # The original script just looped through the whole list, but kept track of where it got to,
+    # to avoid getting stuck on unbuildable crates.
+    #
+    # For now I'm trying to be roughly backwards compatible with the old script's numbers,
+    # but I think that shuffling first is an important bugfixt. Once we are using crates.io sparse
+    # index api without rate-limits, we can think about checking more crates.
+    crates_to_check = random.sample(list(possible_crates), 4 * check_limit)
```
NobodyXu, line 70: I think we should remove already built crates, before sampling?

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,217 @@
...
+    if os.environ.get("RECHECK"):
+        crates_to_check = ["cargo-quickinstall"] + crates_to_check
```
NobodyXu, line 72:
```suggestion
        crates_to_check.append("cargo-quickinstall")
```

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,217 @@
...
+
+        no_default_features = "false"
+        if crate == "gitoxide":
+            features = "max-pure"
+            no_default_features = "true"
+        elif crate == "sccache":
+            features = "vendored-openssl"
+        else:
+            features = ",".join(
+                [feat for feat in version["features"].keys() if "vendored" in feat]
```
NobodyXu, line 93: Nit: could use a generator expression:

```suggestion
                feat for feat in version["features"].keys() if "vendored" in feat
```

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,217 @@
...
+
+def get_popular_crates():
+    with open(
+        Path(__file__).resolve().parent.parent.joinpath("popular-crates.txt"), "r"
+    ) as file:
+        return [
+            line.strip()
+            for line in file
+            if line.strip() != "" and not line.startswith("#")
+        ]
```
NobodyXu, line 155: Something that we could consider using here:

https://github.com/cargo-bins/cargo-quickinstall/issues/268#issuecomment-2329308074

NobodyXu (changes requested):

cronjob_scripts/get_latest_version.py
```diff
@@ -0,0 +1,69 @@
...
+
+def get_index_url(crate: str):
+    """
+    Packages with 1 character names are placed in a directory named 1.
+    Packages with 2 character names are placed in a directory named 2.
+    Packages with 3 character names are placed in the directory 3/{first-character} where {first-character} is the first character of the package name.
+    All other packages are stored in directories named {first-two}/{second-two} where the top directory is the first two characters of the package name, and the next subdirectory is the third and fourth characters of the package name. For example, cargo would be stored in a file named ca/rg/cargo.
+    -- https://doc.rust-lang.org/cargo/reference/registry-index.html#index-files
+    """
+    if len(crate) == 1:
```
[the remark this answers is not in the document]
alsuren: needs to work with 3.8 for now

cronjob_scripts/checkout_worktree.py
```diff
@@ -0,0 +1,51 @@
...
+    This is required for reading the exclude files.
+
+    This is lifted directly from the old trigger-package-build.sh script, and is only expected to
+    work on linux/macos with dash/bash.
+    """
+    worktree_path = f"/tmp/cargo-quickinstall-{target_arch}"
+    bash_script = f"""
+        set -eux
+
+        rm -rf {worktree_path}
```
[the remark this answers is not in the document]
alsuren: temp dir won't work here unless I pull the whole thing inside out and turn it into a context manager. For now, I'm just going to keep the logic how it was.

I don't think shutil buys us anything because these scripts are only ever run on unix machines. I'm assuming that if we have maintainers on windows, they all have access to WSL or similar?

In theory I could use `git ls-files $BRANCH` and `git show $BRANCH:$FILE` to avoid having any checkouts of the "trigger/*" branches at all. I don't think this is worth the bother though, because the builders do a checkout when writing failures, and the symmetry is quite nice.

cronjob_scripts/get_latest_version.py
```diff
@@ -0,0 +1,69 @@
...
+        return f"https://index.crates.io/{crate[:2]}/{crate[2:4]}/{crate}"
+
+
+class CrateVersionDict(TypedDict):
+    name: str
+    vers: str
+    features: dict[str, list[str]]
+
+
+def get_latest_version(crate: str) -> CrateVersionDict:
```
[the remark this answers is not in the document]
alsuren: annoyingly, the sparse index doesn't calculate this for you. I think my logic is correct, and the sparse index api doesn't have a rate limit, so I'd prefer to use it if possible.

I guess we could grep for max_stable_version in the crates.io source code if we wanted to double-check, but I don't think it's the end of the world if we get it wrong.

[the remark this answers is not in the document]
alsuren: in theory, we could make our get_latest_version() in cargo-quickinstall rust client use the crates.io api as well, and keep the logic in sync somehow?

cronjob_scripts/trigger-package-build.py
alsuren: in order to remove already built crates, we need to hammer the crates.io and github releases apis.

The old code kind-of tried to do this, but then it added this strange 4 * CRATE_CHECK_LIMIT cap to the top **before** shuffling and **before** checking for built packages, so it basically ignores anything that's not in the first ~20 results from get-stats.sh (which I think is sorted alphabetically?).

(go read next-unbuilt-package.sh and you'll see what I mean)

When I run this new code, it finds **loads** of crates that the normal cronjob is failing to pick up, so this is definitely better than before.

I think you have more experience with github's rate limits than me, but I say we ship this algorithm for now and tweak it later.

cronjob_scripts/trigger-package-build.py
alsuren: it would be .insert(0, "cargo-quickinstall") - I want to check quickinstall first, before the other random crates, because this is used as part of the release process.

cronjob_scripts/trigger-package-build.py
alsuren: I will add a comment. In practice, that would probably still be put into a text file by some daily cronjob, so the shape of this function probably won't change that much? I haven't really thought about this enough.

cronjob_scripts/trigger-package-build.py
alsuren: Potentially we could, I guess. This would leave us with the case where we make a tag and then die before we can upload the tarball.

Potentially we could check crates.io + git tags for **all** crates, and pick a small random sample to check the github releases for every hour?

This is the kind of proposal that you just wouldn't dream of doing if the whole thing was written in bash, so I feel quite happy that I made this effort.

I'm not sure whether I want to do this in **this** pr. Maybe a fast follow up?

cronjob_scripts/checkout_worktree.py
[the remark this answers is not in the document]
NobodyXu: True, let's keep it for now

cronjob_scripts/get_latest_version.py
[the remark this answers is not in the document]
NobodyXu: I thought the py env in uv is set to 3.10?

Anyway if the py version is too old, then yeah we should skip this for now

cronjob_scripts/trigger-package-build.py
NobodyXu: crates-io dump is updated daily, and I think we could use `actions/cache` to cache it?

cronjob_scripts/trigger-package-build.py
NobodyXu: Yeah we can always add this later

cronjob_scripts/get_latest_version.py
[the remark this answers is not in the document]
NobodyXu: Oh I see you downgraded python to 3.8

cronjob_scripts/get_latest_version.py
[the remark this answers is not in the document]
NobodyXu: Another option would be to use the crates-io dump.

When playing around with it, I discovered that it does have `max_stable_version`

Given that it's updated daily, it should be good enough.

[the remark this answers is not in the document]
alsuren: I would like to have so that we can `make release` and have it immediately run `make recheck` and find the latest version of cargo-quickinstall that we just published, and trigger rebuilds for all architectures.

I **think** that the sparse index protocol gives us this, but I've not tested it.

cronjob_scripts/trigger-package-build.py
alsuren: yeah, possibly, assuming it's not too big. Something for a later PR though.

cronjob_scripts/get_latest_version.py
```diff
@@ -0,0 +1,71 @@
...
+        version
+        for version in versions
+        if not version["yanked"]
+        and not semver.VersionInfo.parse(version["vers"]).prerelease
+    ]
+
+    # FIXME: check that rust actually agrees with semver when it comes to pre-release versions etc.
+    unyanked_versions.sort(key=lambda v: semver.VersionInfo.parse(v["vers"]))
+
+    return unyanked_versions[-1]
```
NobodyXu, line 61: Let's optimize this a bit, the crate might have so many versions that it becomes memory intensive, given that popular crates often have a lot of releases as well:

**And I think you didn't handle the situation where there's no available version to build**, while it's unlikely it could happen, so let's handle that as well.

```suggestion
    max_stable_version = None
    for line in response.text.splitlines():
        version = json.loads(line)
        if version["yanked"]:
            continue
        version = semver.VersionInfo.parse(version["vers"])
       if version.prerelease:
           continue
       if max_stable_version is None or version > max_stable_version:
           max_stable_version = version

    return max_stable_version 
```

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,184 @@
...
+    https://github.com/cargo-bins/cargo-quickinstall/issues/268#issuecomment-2329308074),
+    but probably not every time we run this script. We will need to think about this.
+    """
+    with open(
+        Path(__file__).resolve().parent.parent.joinpath("popular-crates.txt"), "r"
+    ) as file:
+        return [
+            line.strip()
+            for line in file
+            if line.strip() != "" and not line.startswith("#")
```
NobodyXu, line 121: Fixing a bug, where a `"[[:space:]]+#"` is treated as a crate:

```suggestion
            stripped
            for line in file
            if (stripped := line.strip()) and not stripped.startswith("#")
```

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,184 @@
...
+        ]
+
+
+def get_excludes(tracking_worktree_path: str, days: int, max_failures: int):
+    """
+    if a crate has reached `max_failures` failures in the last `days` days then we exclude it
+    """
+    failures = Counter()
+
+    failure_filenames = list(Path(tracking_worktree_path).glob("*-*-*"))
```
NobodyXu, line 131: For these files, I think we don't have any mechanism to delete them?

So it'd keep growing, I'd recon we need to add a cleanup mechanism in another PR, otherwise this would get inefficient eventually.

cronjob_scripts/trigger-package-build.py
NobodyXu: Think it's simple enough, was polars I was able to do it around 10 lines, and it's quite efficient (in the linked issue)

NobodyXu (changes requested):

cronjob_scripts/trigger-package-build.py
NobodyXu: > I'm not sure whether I want to do this in this pr. Maybe a fast follow up?

Yeah a follow-up is 👍

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,217 @@
...
+
+    if target_arch == "all":
+        return list(TARGET_ARCH_TO_BUILD_OS.keys())
+
+    try:
+        rustc_version_output = subprocess.run(
+            ["rustc", "--version", "--verbose"], capture_output=True, text=True
+        )
+        current_arch = [
+            line.split("host: ")[1]
```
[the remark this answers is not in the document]
NobodyXu: cc @alsuren in case you missed this

cronjob_scripts/trigger-package-build.py
NobodyXu: cc @alsuren in case you missed this

cronjob_scripts/get_latest_version.py
alsuren: I suspect that the RAM usage from parsing the sparse crates.io index will always be dwarfed by the background RAM usage of the python interpreter + libs. Your way does feel a bit cleaner though, and makes the empty result case more obviously correct, so I will refactor it into something more like that shape.

cronjob_scripts/trigger-package-build.py
alsuren: backlogged as https://github.com/cargo-bins/cargo-quickinstall/issues/276

cronjob_scripts/trigger-package-build.py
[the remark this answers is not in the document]
alsuren: I fixed it when I moved it into architectures.py

cronjob_scripts/trigger-package-build.py
alsuren: cheers. I had a bit of clean up of my blanked except: blocks just now, so hopefully it's a bit cleaner now.

cronjob_scripts/trigger-package-build.py
alsuren: Ah. I thought about this again, and I don't think we can use git tags:

We used to make tags like `cargo-benchcmp-0.4.5-aarch64-apple-darwin` and add a single tarball to it. We now make tags like `cargo-benchcmp-0.4.5` and add a tarball per arch.

This means that we can't just read the output of `git tag` to determine if we have built a thing for a particular arch, because it might have been built for a different architecture.

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,182 @@
...
+def get_current_version_if_unbuilt(
+    repo_url: str,
+    crate: str,
+    target_arch: str,
+) -> CrateVersionDict | None:
+    version = get_latest_version(crate)
+    if not version:
+        return None
+
+    url = f"{repo_url}/releases/download/{crate}-{version['vers']}/{crate}-{version['vers']}-{target_arch}.tar.gz"
```
NobodyXu, line 171: I think using a python library for it would be more readable https://pypi.org/project/PyGithub/

For the rate-limit problem, I think we can provide a read-only GitHub token via secrets.

While `secrets.GITHUB_TOKEN` would work, it
 - has a rate limit of 1000 requests per hour
 - has too many permissions within the repo

If we create a read-only token using a robot account (which I've setup), we will have:
 - 5000 requests per hour
 - has read-only permission and cannot access any private info

cronjob_scripts/trigger-package-build.py
```diff
@@ -0,0 +1,182 @@
...
+
+    # check random crates up to a limit.
+    check_limit = int(os.environ.get("CRATE_CHECK_LIMIT", 5))
+    # Note that the old next-unbuilt-package.sh script actually checked the top 4 * CRATE_CHECK_LIMIT
+    # crates from get-stats.sh + popular-crates.txt rather than shuffling the whole lot before checking.
+    # The original script just looped through the whole list, but kept track of where it got to,
+    # to avoid getting stuck on unbuildable crates.
+    #
+    # For now I'm trying to be roughly backwards compatible with the old script's numbers,
+    # but I think that shuffling first is an important bugfixt. Once we are using crates.io sparse
```
NobodyXu, line 57:
```suggestion
    # but I think that shuffling first is an important bugfix. Once we are using crates.io sparse
```

cronjob_scripts/architectures.py
```diff
@@ -0,0 +1,50 @@
...
+
+    try:
+        current_arch = [
+            line.removeprefix("host: ")
+            for line in rustc_version_output.stdout.splitlines()
+            if line.startswith("host: ")
+        ][0]
+        return [current_arch]
+    except IndexError:
+        raise ValueError(f"rustc did not tell us its host: {rustc_version_output}")
```
NobodyXu, line 50: - strip the line before processing it
 - use `if not` instead of exception handling (I personally don't like exception very much, but if you think it's ok I'd be fine about it).

```suggestion
    current_target = [
        stripped.removeprefix("host: ")
        for line in rustc_version_output.stdout.splitlines()
        if (stripped := line.strip())
        if stripped.startswith("host: ")
    ]
    if not current_target:
        raise ValueError(f"rustc did not tell us its host: {rustc_version_output}")
    return current_target
```

NobodyXu, line 50: - strip the line before processing it
 - use `if not` instead of exception handling (I personally don't like exception very much, but if you think it's ok I'd be fine about it).

```suggestion
    current_target = [
        stripped.removeprefix("host: ")
        for line in rustc_version_output.stdout.splitlines()
        if (stripped := line.strip())
        if stripped.startswith("host: ")
    ]
    if not current_target:
        raise ValueError(f"rustc did not tell us its host: {rustc_version_output}")
    return current_target
```

NobodyXu (changes requested):

cronjob_scripts/trigger-package-build.py
NobodyXu: @alsuren if you think this is too much work for this PR, we could leave this to a follow-up PR.

alsuren: My main goals in this PR are to stop using the old stats server and to delete 400 lines of bash, so I am going to leave it like this.

I also quite like the symmetry of using exactly the same url that the quickinstall client uses. The old code even used the same http client (`curl`) but I decided that the connection pooling from requests was too much of a win.

> We could use the github api to get all artifacts and cache it.

I'm a bit wary of using a "do a lot of work and rely on caches to save us" approach. I've so far been taking the approach of "do as little work as possible and still make forward progress". I realise that there is a cutover though. As we check more crates every hour, we will want to have some kind of cached pre-filter to reduce the number of http requests we need to make. We would need to be wary of stale caches as well: if the cache says the package is there then we can trust that, but if the cache says it's not there, we need to re-check to make sure we didn't build it in the last hour.

In another unrelated note: I think it would be really interesting to gather stats around our build processes as well as our client requests. This would allow us to make better decisions when we make tweaks to the scheduling algorithm. I'm quite annoyed that the new influxdb cloud org doesn't have good notebook/dashboards. I really don't want to have to plug it into grafana just to share some graphs.

NobodyXu (approved):

NobodyXu:
Thank you, it LGTM, let's merge this in and switch to influx!

i will update the cargo-binstall to use the influx API afterwards and cuts a new release

alsuren:
We also need to get https://github.com/cargo-bins/cargo-quickinstall/pull/165 merged, since that's what's currently powering the influx stuff (note that the vercel stats server is already forwarding stats to the new server, so as long as we get this done in the next month, it should't make much difference).

NobodyXu:
> not sure where my print() statements output is disappearing to in the github actions output :(

For output, we could use grouped output which is more readable

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#grouping-log-lines

so we can have a 

```python
@contextmanager
def group_output(title):
    print("::group::{title}")
    yield
    print("::endgroup::")

with group_output("Collecting popular crates"):
    print(...)
```

will then looks like image

Issue #282, opened 2024-09-10
Bump py ver to 3.11/3.12?


NobodyXu:
https://github.com/cargo-bins/cargo-quickinstall/actions/runs/10784214724/job/29907455152#step:3:45

cronjob fails due to type annotation, it seems like python 3.8 wasn't able to handle type annotation and ignore the, cc @alsuren

PR #283, opened 2024-09-10
Fix python

alsuren:
python 3.8 doesn't understand `list[str]` as a type annotation at runtime (see error in https://github.com/cargo-bins/cargo-quickinstall/issues/282).

This wasn't caught because I was accidentally smoke-testing locally on a more modern python.

I was also pushing to the actions branch to test on CI, but forgot to do that between making a last minute refactor and merging to main.

I think this test will avoid similar problems with shipping broken python code.

Note that we don't do any mypy type checking in CI, so it's up to whatever you have enabled in your IDE to catch any discrepancies.


NobodyXu (approved):
Thanks, LGTM, test failure due to cargo-quickinstall v0.3.0 not having pre-built

Issue #284, opened 2024-09-10
Re-think scheduling of builds


alsuren:
Currently we do:

* for each arch in series:
  * pick randomly from the popular-crates.txt + the list of requested crates for that arch
    * check and trigger builds for that package

Here is a slightly convoluted scheme that might allow us to fairly make progress while also prioritising checking of things that are likely to be useful to people

* construct the following list of crates (n lists for each arch):
  * for each arch:
    * the list of install requests with the "built-from-source" status in the last day (ordered by popularity, including versions - WARNING: this is very likely to result in the same crate being built many times in parallel if someone runs the cronjob manually on ci/locally)
    * the list of install requests with the "built-from-source" status in the last day (shuffled, including versions)
    * the list of install requests with any other status in the last day (shuffled, with versions stripped)
    * popular-crates.txt (shuffled)
    * a list with just cargo-quickinstall and nothing else
* shuffle the list of lists and then round-robin between each list:
  * take the head of the list
  * if we already checked this package for this arch, skip it
  * if we already tried to build this package 5 times in the last 7 days, skip it (TODO: also record the version in the failure log?)
  * check its latest version + if we have a package built for it. If we don't have a package built for it, trigger a build
    * if we already triggered m package builds since the start of the run, exit (probably set this to n times the number of arches?)
* if we hit a github rate limit when doing this, this is fine and we just bomb out having made some progress?
* if we get to the end without triggering any builds then that's fine too: there's nothing to do.

Questions:

1) would there be any value in including cargo-binstall's tree of fallback architectures here?
2) can we make this simpler while keeping it fair and resilient to rate limit problems
3) can we gain anything by doing things on a crate-by-crate basis (so we make the most of any crates.io index fetches we make)?

NobodyXu:
> WARNING: this is very likely to result in the same crate being built many times in parallel if someone runs the cronjob manually on ci/locally

We have a mechanism to cancel duplicate builds, though it does waste machine time as the previous in progress job is just cancelled.

alsuren:
ah actually I didn't think of that. Seems we have our `concurrency` set correctly:

```
concurrency:
  group: build-package-${{ github.event.inputs.crate }}-${{ github.event.inputs.version }}-${{ github.event.inputs.target_arch }}
```

[...]

PR #285, opened 2024-09-10
Get popular crates from crates.io db dump

NobodyXu:


NobodyXu:
Turns out that parsing popular crates from crates.io db dump would only take around 12-14s in CI, polars is so damn efficient

cronjob_scripts/trigger-package-build.py
```diff
@@ -41,10 +42,10 @@ def trigger_for_arch(target_arch: str):
     tracking_worktree_path = checkout_worktree_for_arch(target_arch)
     excludes = get_excludes(tracking_worktree_path, days=7, max_failures=5)
 
-    popular_crates = get_popular_crates()
-    requested_crates = get_requested_crates(period="1 day", arch=target_arch)
+    possible_crates = set(get_crates_io_popular_crates())
```
alsuren, line 45: this now downloads 700MB from crates.io every time we want to check crates (every hour + anytime I want to smoke test anything locally), with no cache.

I was thinking about this the other day, and we could potentially only extract the files we're planning to ingest, and save it to somewhere in /tmp, with a cache key based on today's date or something?

alsuren: sorry... once per arch because there get_crates_io_popular_crates is not wrapped in @functools.lru_cache

cronjob_scripts/crates_io_popular_crates.py
```diff
@@ -0,0 +1,83 @@
...
+    response.raise_for_status()
+
+    with tarfile.open(fileobj=response.raw, mode='r:gz') as tarball:
+        while True:
+            member = tarball.next()
+            if member is None:
+                return
+            yield TarEntry(tarball, member)
+
+def get_crates_io_popular_crates(minimum_downloads=4000):
```
alsuren, line 38: I have no idea how many crates this will return. If it returns a lot of crates then our current scheme will take an awfully long time getting around to checking the most popular ones. Can we limit the total number to 100 or something? (not sure what a sensible number should be). See #284 

I was also thinking about whether we could pick a sampling algorithm that favours highly requested packages but still also checks rarely requested packages sometimes. Maybe multiply up each crate in the list by its request count, and then shuffle, and then deduplicate? There's probably a better way to do it though.

NobodyXu: We can tune the packages returned via `minimum_downloads`.

If the minimum_downloads is 4000, then it would return 929 crates.

If you increase it, then the number decrease very quickly.

You can play with that file a bit

> I was also thinking about whether we could pick a sampling algorithm that favours highly requested packages but still also checks rarely requested packages sometimes. 

IMO that is covered by our stats server.

cronjob_scripts/trigger-package-build.py
NobodyXu: Yes according to [polars user guide](https://docs.pola.rs/user-guide/io/parquet/#)

> Loading or writing [Parquet files](https://parquet.apache.org/) is lightning fast as the layout of data in a Polars DataFrame in memory mirrors the layout of a Parquet file on disk in many respects.

And there's an API for doing it, with optional compression.

https://docs.pola.rs/api/python/stable/reference/api/polars.DataFrame.write_parquet.html#polars.DataFrame.write_parquet

If we are going to cache stuff and current progress in trigger/* branch, then this will not be necessary.

cronjob_scripts/crates_io_popular_crates.py
NobodyXu: I've increased `minimum_downloads` to `200000`, so it only returns ~300 crates

cronjob_scripts/trigger-package-build.py
NobodyXu: I've added caching, the file to be cached is only 2785 B large, much smaller than the 700MB file @alsuren

cronjob_scripts/crates_io_popular_crates.py
```diff
@@ -0,0 +1,97 @@
...
+        df = pl.scan_parquet(cached_path).collect(streaming=True)
+    else:
+        df = (
+            get_crates_io_popular_crates_inner(minimum_downloads)
+            # TODO: Use streaming, maybe use sink_parquet instead?
+            # https://github.com/pola-rs/polars/issues/18684
+            .collect()
+        )
+        df.write_parquet(cached_path, statistics=False)
+    return (row[0] for row in df.iter_rows())
```
alsuren, line 83: I was wondering if we could share anything like my decorator at https://github.com/cargo-bins/cargo-quickinstall/pull/288/files#diff-5cb8ada8e1547a375244dba018abfe68cdee554ee84f24b6ee342b050878c3a0R105-R128



It feels like you're aggressively optimising for speed here. Remember that we have a sleep(30) in the inner loop when we are triggering packages, so speed probably doesn't matter **that** much.

I suspect you'd do better if you return the dataframe, and let the caller do `set(df["name"])`  or something(polars is a based columnar datastore, so getting the column as a series will probably be quicker). It might be nicer for the downstream code to work in terms of the polars.Series rather than converting to an iterator and collecting into a set(). I'm trying to avoid getting caught up micro-optimising this, because I think simplicity is best for the cronjob code.

Also remember that we will be calling this function once per architecture, so maybe it's best to just keep the dataframe in memory (I will probably be fixing that soon though, so maybe don't bother)

cronjob_scripts/crates_io_popular_crates.py
```diff
@@ -0,0 +1,97 @@
...
+
+    with tarfile.open(fileobj=response.raw, mode='r:gz') as tarball:
+        while True:
+            member = tarball.next()
+            if member is None:
+                return
+            yield TarEntry(tarball, member)
+
+def get_crates_io_popular_crates_inner(minimum_downloads=200000):
+    files_to_extract = ("crate_downloads.csv", "crates.csv", "default_versions.csv", "versions.csv")
```
alsuren, line 42:
```suggestion
    files_to_extract = ["crate_downloads.csv", "crates.csv", "default_versions.csv", "versions.csv"]
``` 

use list rather than tuple if you have uniform types and the length is not relevant - you would use &[&str] in rust, so list in python

cronjob_scripts/crates_io_popular_crates.py
```diff
@@ -0,0 +1,97 @@
...
+    files_extracted = 0
+    dfs = {}
+    for entry in download_tar_gz('https://static.crates.io/db-dump.tar.gz'):
+        name = entry.is_one_of_csvs_interested(files_to_extract)
+        if name:
+            dfs[name] = pl.scan_csv(entry.get_file_stream())
+            files_extracted += 1
+            if files_extracted == len(files_to_extract):
```
alsuren, line 50: dealing with tarballs is a bit funky because they can have state/seeking, so I understand why you want to write a helper for it. I would probably skip the TarEntry class and instead split the responsibilities into:
```
def get_dataframes(download_url: str, filenames: list[str]) -> dict[str, pd.LazyFrame]:
    ... # inline the class stuff here, return dfs

def get_crates_io_popular_crates_inner(minimum_downloads= 200_000):
    dfs = get_dataframes(files_to_extract)
    return ( ... # everything below this line
```

(keep the class if you want, but definitely split this into two functions, so there is a clear separation of responsibilities)

cronjob_scripts/crates_io_popular_crates.py
alsuren: ... the parquet file is 2785B. We should definitely just be calling read_parquet() and returning the df rather than even thinking about being clever about it.

cronjob_scripts/crates_io_popular_crates.py
```diff
@@ -0,0 +1,97 @@
...
+                    .filter(pl.col("bin_names") != "{}", pl.col("yanked") == "f")
+                    .filter(pl.col("downloads") > minimum_downloads)
+                    .select("name")
+                )
+
+    raise RuntimeError(f"Failed to find all csvs {files_to_extract}")
+
+
+def get_crates_io_popular_crates(minimum_downloads=200000):
+    cached_path = Path("cached_crates_io_popular_crates.parquet")
```
alsuren, line 69: add this to .gitignore or move it to /tmp?

alsuren (commented):
I actually quite like it in this shape.

Let's stop micro-optimising though: the on-disk cache of a 2kB file (less than a single disk block) is the headline optimisation that makes this thing fast. Everything else is just mental overhead.

alsuren (approved):

cronjob_scripts/crates_io_popular_crates.py
```diff
@@ -0,0 +1,97 @@
...
+
+def download_tar_gz(url):
+    response = requests.get(url, stream=True)
+    response.raise_for_status()
+
+    with tarfile.open(fileobj=response.raw, mode='r:gz') as tarball:
+        while True:
+            member = tarball.next()
+            if member is None:
+                return
```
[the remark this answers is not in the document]
NobodyXu: Thanks, [official doc for `tarfile`](https://docs.python.org/3/library/tarfile.html) does not specify if it's an iterable or not, but when looking at it using `help(tarfile.TarFile)` it's definitely iterable/

cronjob_scripts/crates_io_popular_crates.py
```diff
@@ -0,0 +1,97 @@
...
+    raise RuntimeError(f"Failed to find all csvs {files_to_extract}")
+
+
+def get_crates_io_popular_crates(minimum_downloads=200000):
+    cached_path = Path("cached_crates_io_popular_crates.parquet")
+    if cached_path.is_file():
+        # TODO: Once `iter_rows()` can be used on LazyFrame, use it for better perf
+        #  and less ram usage.
+        # https://github.com/pola-rs/polars/issues/10683
+        df = pl.scan_parquet(cached_path).collect(streaming=True)
```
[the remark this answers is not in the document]
NobodyXu: Yes, it's the same here

cronjob_scripts/crates_io_popular_crates.py
NobodyXu: I return an iterable here since the the previous `get_popular_crates()` return a `list`.

Returning a `DataFrame` is ok, but it doesn't make much sense for the caller, it just expects something that is iterable and can be passed to `set`

```python3
>>> pl.DataFrame({"name": ["a"]})
shape: (1, 1)
┌──────┐
│ name │
│ ---  │
│ str  │
╞══════╡
│ a    │
└──────┘
>>> df = pl.DataFrame({"name": ["a"]})
>>> list(row for row in df)
[shape: (1,)
Series: 'name' [str]
[
	"a"
]]
```

`pl.DataFrame` is not iterable`, `pl.DataFrame.iter_rows` returns `list[tuple[str]]`:

```python3
>>> pl.DataFrame({"name": ["a"]})
shape: (1, 1)
┌──────┐
│ name │
│ ---  │
│ str  │
╞══════╡
│ a    │
└──────┘
>>> df = pl.DataFrame({"name": ["a"]})
>>> list(row for row in df)
[shape: (1,)
Series: 'name' [str]
[
	"a"
]]
```

so it's better to return a generator of `Iterable[str]`.

cronjob_scripts/crates_io_popular_crates.py
NobodyXu: Thanks updated

cronjob_scripts/crates_io_popular_crates.py
NobodyXu: Thanks, updated and it does look better

cronjob_scripts/crates_io_popular_crates.py
NobodyXu: Added this to `.gitignore`

.github/actions/cronjob-setup/action.yml
```diff
@@ -0,0 +1,20 @@
...
+  using: composite
+  steps:
+    - uses: actions/setup-python@v5
+      with:
+        python-version-file: "cronjob_scripts/.python-version"
+        cache: "pip"
+        cache-dependency-path: "**/requirements.txt"
+
+    - name: Set current date as env variable
+      run: echo "NOW=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
NobodyXu: Deleted `popular-crates.txt`, I think the out-of-date would be ok in practice, as it seems to be relatively stable.

And given that how slow the rebuild is, I think being a bit outdated is ok.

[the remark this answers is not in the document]
NobodyXu: If that turns out to be a problem, we could use UTC time or manually -2h

PR #293, opened 2024-09-13
Use latest zig-cc compiler

### LINKED ISSUE #250: Broken build for jj-cli
[related issue in the jj repo](https://github.com/martinvonz/jj/issues/3844)

jj provides [prebuilt binaries itself](https://github.com/martinvonz/jj/releases/tag/v0.18.0), but only statically linked against musl on linux. Which is fine, because that works on any linux system, whatever the system libc is. However, `cargo binstall jj-cli` fell back to the (borked) binary provided by quickinstall. I did not notice this at first, even though it does say so in the output of `cargo-binstall`.

I'm not sure what the right thing to do here is, but having `cargo-binstall` install from a different source than the official one, even though the official one does in fact provide binaries, is incredibly confusing.

NobodyXu:
Fixed #250

### DIFF
--- a/zigbuild-requirements.txt
+++ b/zigbuild-requirements.txt
@@ -2,4 +2,4 @@
 cargo-zigbuild
 
 ###### Requirements with Version Specifiers ######
-ziglang < 0.11 # zig 0.11 causes link failure in our CI
+ziglang


alsuren (approved):
I will defer to your judgement on this one.

PR #294, opened 2024-09-13
Enable all feature including "bundled"

NobodyXu:
Motivation:

Some crates, e.g. https://docs.rs/crate/cargo-docset/latest/features , have a `bundled-*` feature to build a vendored C/C++ library, which is necessary for cross-compilation and reliable distribution.

### DIFF
--- a/cronjob_scripts/trigger-package-build.py
+++ b/cronjob_scripts/trigger-package-build.py
@@ -80,7 +80,7 @@ def trigger_for_arch(target_arch: str):
             features = "vendored-openssl"
         else:
             features = ",".join(
-                feat for feat in version["features"].keys() if "vendored" in feat or bundled in feat
+                feat for feat in version["features"].keys() if "vendored" in feat or "bundled" in feat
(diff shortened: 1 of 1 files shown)


cronjob_scripts/trigger-package-build.py
```diff
@@ -80,7 +80,7 @@ def trigger_for_arch(target_arch: str):
             features = "vendored-openssl"
         else:
             features = ",".join(
-                feat for feat in version["features"].keys() if "vendored" in feat
+                feat for feat in version["features"].keys() if "vendored" in feat or bundled in feat
```
alsuren, line 83:
```suggestion
                feat for feat in version["features"].keys() if "vendored" in feat or "bundled" in feat
```

I feel like Ruff would have caught this.

NobodyXu: Oops thanks.

I think maybe we want mypy/ruff/ruff-format in CI?

PR #300, opened 2024-09-26
List of lists approach

alsuren:
~This is a bit of a mess: I started pulling at various threads and ended up doing a whole bunch of cleanup work. Once I'm happy with the end-result, I can split it into some more sensible PRs or we can just review it as-is.~

~I probably won't have time to look at this again until next week.~

I had a bit of a git history cleanup since you last looked, and backed out the target_arch -> target rename because it was too distracting.

Running on the `actions` branch to see how it performs. It seems to be triggering a lot of builds, so I tuned it down a bit.


add-exclude.sh
```diff
@@ -1,22 +1,162 @@
...
+
+if [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -I --output /dev/null "${REPO}/releases/download/${CRATE}-${VERSION}/${CRATE}-${VERSION}-${TARGET}.tar.gz"; then
+    echo "${CRATE}/${VERSION}/${CRATE}-${VERSION}-${TARGET}.tar.gz already uploaded. Skipping."
+    exit 0
+fi
+
+# Install llvm
+if [ "${RUNNER_OS?}" == "Windows" ]; then
+    choco install llvm
+    elif [ "${RUNNER_OS?}" == "Linux" ]; then
```
NobodyXu, line 58: Nit: extra indent for this elif


```suggestion  
elif [ "${RUNNER_OS?}" == "Linux" ]; then
```

cronjob-scripts/cronjob_scripts/stats.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    global _influxdb_client
+    if _influxdb_client is None:
+        _influxdb_client = InfluxDBClient3(
+            host=HOST, token=TOKEN, org=ORG, database=DATABASE
+        )
+
+    # Old clients (going via the old stats server) don't report status, so we filter them out.
+    # FIXME: seems like there are some entries of the form:
+    # {"agent": "binstalk-fetchers/0.10.0", "status": "start", ... }
```
NobodyXu, line 38: AFAIK [binswap-github](https://crates.io/crates/binswap-github) is using an old version of binstalk.

I previously propose to move it into cargo-binstall but we were too busy and forgot it

cronjob-scripts/cronjob_scripts/stats.py
```diff
@@ -0,0 +1,71 @@
...
+
+def get_requested_crates(period: str, target: str | None) -> list[CrateAndVersion]:
+    df = get_stats(period=period)
+
+    if target is not None:
+        # for now, if target in influxdb is null then build for all targets
+        # (just until we have a day's worth of data with `target` set)
+        mask = (df["target"] == target) | df["target"].is_null()
+
+        df = df.filter(mask)
```
NobodyXu, line 69: I think it'd be more ergonomic to use the following code, and perhaps more efficient (from my memory of pandas, `df["target"] == target ` generates a boolean array).

```suggestion
        df = df.filter((pl.col("target") == target) | (pl.col("target").is_null()))
```

cronjob-scripts/cronjob_scripts/stats.py
```diff
@@ -0,0 +1,71 @@
...
+
+    return df[["crate", "version"]].to_dicts()
```
NobodyXu, line 71: I wonder if we should just export the polars.DataFrame directly?

NobodyXu (changes requested):

cronjob-scripts/cronjob_scripts/trigger_package_build.py
```diff
@@ -0,0 +1,196 @@
...
+    requested_version: str | None = None,
+) -> CrateVersionDict | None:
+    
+    # FIXME(probably never): in theory we could skip a round-trip to crates.io if we already know the version,
+    # but the code ends up looking ugly so I decided not to bother.
+    version = get_latest_version(crate, requested_version)
+    if not version:
+        return None
+
+    url = f"{repo_url}/releases/download/{crate}-{version['vers']}/{crate}-{version['vers']}-{target}.tar.gz"
```
NobodyXu, line 189: I think we might want to provide github token, shall save us a lot headache from rate limiting

NobodyXu (changes requested):

cronjob-scripts/cronjob_scripts/stats.py
NobodyXu: Closed, looking at how we use it, it'd be a bit hard.

add-exclude.sh
```diff
@@ -1,22 +1,162 @@
 #!/bin/bash
 set -euxo pipefail
 
-if [ $# != 1 ]; then
-    echo "Usage: $0 /path/to/trigger/branch"
```
NobodyXu, line 5: I think you have accidentally copied over build-version.sh into this file?

NobodyXu (changes requested):

build-version.sh
```diff
@@ -32,6 +32,146 @@ curl_slowly() {
...
+    
+    export CARGO=cargo-zigbuild
+    # Use our own pkg-config that fails for any input, since we cannot use
+    # locally installed lib in cross-compilation.
+    export PKG_CONFIG="$PWD/pkg-config-cross.sh"
+}
+
+REPO="$(./get-repo.sh)"
+
+if [ "${ALWAYS_BUILD:-}" != 1 ] && curl_slowly --fail -I --output /dev/null "${REPO}/releases/download/${CRATE}-${VERSION}/${CRATE}-${VERSION}-${TARGET}.tar.gz"; then
```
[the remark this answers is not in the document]
alsuren: I backed this change out during my rebase. I don't know if we can use `gh` in the build-version script so I won't pick this up right now. I might do the other one though.

cronjob-scripts/cronjob_scripts/trigger_package_build.py
alsuren: I switched this to use `gh release view $tag --json=assets` and catch "release not found"

cronjob_scripts/trigger_package_build.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # we shuffle the list of popular crates from crates.io once, and use it for every arch, to slightly
+    # reduce the number of requests we need to make to crates.io. It probably isn't worth it though.
+    popular_crates = get_crates_io_popular_crates()
+    random.shuffle(popular_crates)
+
+    for target in targets:
+        queue = get_requested_crates(period="1 day", target=target)
+        random.shuffle(queue)
+        queues.append(("requested", target, cast(List[CrateAndMaybeVersion], queue)))
```
alsuren, line 68: This cast is kind-of horrible. I forgot that list[A] is not a subtype of list[B] even if A is a subtype of B (contravariance or something).

Even more ugly because cast() just ignores type of input.

I don't really care enough though.

cronjob_scripts/stats.py
```diff
@@ -12,40 +17,65 @@
...
+    ask for exactly the architectures we want, but that would require us to make the whole script
+    async, and we know that cronjob will eventually ask for all architectures anyway.
+    """
     global _influxdb_client
     if _influxdb_client is None:
         _influxdb_client = InfluxDBClient3(
             host=HOST, token=TOKEN, org=ORG, database=DATABASE
         )
 
+    # Old clients (going via the old stats server) don't report status, so we filter them out.
```
NobodyXu, line 36: I think keeping data from old client is ok?

NobodyXu (changes requested):

cronjob_scripts/stats.py
alsuren: yeah, fair point: phasing out the old clients seems to be taking longer than I expected, according to https://alsuren.grafana.net/d/cdx9qygf9e0owe/cargo-quickinstall?from=now-7d&to=now&timezone=browser&viewPanel=panel-3 . I will have another think about this.

cronjob_scripts/stats.py
```diff
@@ -12,40 +17,71 @@
...
+    # get phased out and then this method will only return crate versions which ended up installing
+    # from source.
+    #
+    # FIXME: seems like there are some entries of the form:
+    # {"agent": "binstalk-fetchers/0.10.0", "status": "start", ... }
+    # Apparently binswap-github is using an old version of binstalk. See:
+    # https://github.com/cargo-bins/cargo-quickinstall/pull/300/files#r1778063083
     query = """
-        SELECT DISTINCT crate
+        SELECT DISTINCT crate, target, version, status
```
NobodyXu, line 46: I feel like `DISTINCT ON` is what we actually want, but not sure if it is supported

```suggestion
        SELECT DISTINCT ON(crate, target) crate, target, version, status
```

cronjob_scripts/trigger_package_build.py
```diff
@@ -0,0 +1,249 @@
...
+
+
+def main():
+    """
+    Triggers a package build in a vaguely fair way.
+
+    General approach:
+    * construct the following list of crates (n lists for each target):
+        * for each target:
+            * TODO (probably never): the list of install requests with the "built-from-source" or "not-found" status in the last day (ordered by popularity, including versions - WARNING: this is very likely to result in the same crate being built many times in parallel if someone runs the cronjob manually on ci/locally)
```
NobodyXu, line 32: This comment is a bit too long.

Can you split it into multiple lines?

cronjob_scripts/trigger_package_build.py
```diff
@@ -0,0 +1,249 @@
...
+        exit(1)
+
+    targets = get_target_architectures()
+    # FIXME: make a better type than tuple here.
+    # We mean list of (type, target, list of crates and versions)
+    queues: list[tuple[str, str, list[CrateAndMaybeVersion]]] = []
+
+    # we shuffle the list of popular crates from crates.io once, and use it for every arch, to slightly
+    # reduce the number of requests we need to make to crates.io. It probably isn't worth it though.
+    popular_crates = get_crates_io_popular_crates()
```
NobodyXu, line 65: Can we filter out excluded crate here?

cronjob_scripts/trigger_package_build.py
```diff
@@ -0,0 +1,249 @@
...
+        queues.append(
+            ("requested", target, cast(List[CrateAndMaybeVersion], all_requested))
+        )
+
+        # This should be a small list of crates, so we should get around to checking everything here
+        # pretty often. This doesn't include failures from old clients (via the old stats server)
+        # because they don't report status.
+        failed = get_requested_crates(
+            period="1 day", target=target, statuses=["built-from-source", "not-found"]
+        )
```
NobodyXu, line 82: Can we filter it out excluded crate early here?

NobodyXu (changes requested):

cronjob_scripts/stats.py
alsuren: I think status and version will both be very low cardinality (when grouped by (crate, target) for the last day). 

Generally we always ask for the latest version, so there will be either 1 or 2 different values of `version` per crate, depending on whether there was a release today.

If a crate version is unbuilt then we will receive some status=null spam from old clients and status=not-found from new clients. I think this will result in no more than twice the rows back from influxdb compared to your method, so I don't think it's worth optimising unless influxdb starts rate limiting us (which I am not expecting to happen if we make this query once per hour). We need to keep both statuses separate so that we can construct the 'not-found' high priority short list.

NobodyXu: Doesn't `DISTINCT crate` remove all rows that have the same crate name?

I.e. if two rows have different target names but the same crate name, only one of them will be kept

cronjob_scripts/trigger_package_build.py
```diff
@@ -0,0 +1,285 @@
...
+    failures: Counter[str] = Counter()
+
+    failure_filenames = list(Path(tracking_worktree_path).glob("*-*-*"))
+    failure_filenames.sort(reverse=True)
+
+    for failure_filename in failure_filenames[:days]:
+        with open(failure_filename, "r") as file:
+            failures.update(
+                Counter(stripped for line in file if (stripped := line.strip()))
+            )
```
NobodyXu, line 222:
```suggestion
@lru_cache
def get_repo_url() -> str:
```

NobodyXu (changes requested):

cronjob_scripts/stats.py
alsuren: in  postgres no: https://www.postgresql.org/docs/current/sql-select.html#SQL-DISTINCT

        SELECT DISTINCT crate, target, version, status

would be equivalent to

        SELECT DISTINCT on(crate, target, version, status) crate, target, version, status

NobodyXu: Thanks, if this is the case then I'm ok with this SQL code

NobodyXu (approved):
I think it is all good

PR #325, opened 2024-11-12
recheck needs github token

alsuren:
see https://github.com/cargo-bins/cargo-quickinstall/actions/runs/11793818866/job/32850152585#step:6:93

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -30,4 +30,4 @@ jobs:
       - name: Trigger tarball builds of cargo-quickinstall
         run: make recheck-self-only
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-plz.yml
```diff
@@ -29,3 +29,5 @@ jobs:
       - uses: ./.github/actions/cronjob-setup
       - name: Trigger tarball builds of cargo-quickinstall
         run: make recheck-self-only
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
NobodyXu, line 33:
```suggestion
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

```
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:\n  env:\n    GH_TOKEN: ${{ github.token }}
```

NobodyXu (changes requested):

.github/workflows/release-plz.yml
alsuren: interesting. I just copy-pasted that from somewhere else and it seems to work. I guess the cronjob uses GH_TOKEN so go for it

NobodyXu (approved):

PR #364, opened 2025-05-08
Correct example of how to use in CI in README

platy:
It looks like the quickinstall behaviour was changed to call out to cargobinstall if available, but binstall doesn't produce the output suggested


NobodyXu (commented):
Thanks for the updated expanded command!

I'm not sure if disabling the binstall in CI example is great, maybe we can document that (both with and without binstall)?

NobodyXu:
Aha that's true.

NobodyXu:
Thanks!

PR #402, opened 2025-07-09
Fix cargo-install caching in selfbuild.yml not working on windows

NobodyXu:


.github/actions/cache-cargo-install/action.yml
```diff
@@ -0,0 +1,19 @@
+name: Setup env for cronjob and cache
```
NobodyXu, line 1:
```suggestion
name: Cache cargo-install target dir
```

.github/actions/cronjob-setup/action.yml
```diff
@@ -1,4 +1,4 @@
-name: Setup env for cronjob and cache
+name: Cache cargo-install target dir
```
NobodyXu, line 1:
```suggestion
name: Setup env for cronjob and cache
```

PR #416, opened 2025-07-30
chore(CI): customize build workflow's run-name

### LINKED ISSUE #415: display built package name in workflow run titles
Searching the releases for a specific package/version is pretty intuitive. But finding which CI workflow run that (successfully or not) uploaded the package to releases is much harder. This is because GitHub does not support searching logs of workflow runs.

## Proposal
Simply add a customized title to the [.github/workflows/build-package.yml](https://github.com/cargo-bins/cargo-quickinstall/blob/main/.github/workflows/build-package.yml) using [`run-name`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#run-name):

```yml
run-name: Build Package ${{ inputs.crate }} ${{ inputs.version }} (${{ inputs.build_os }})
```
This would add enough information for searching workflow runs about a specific build.

---
PS: I also think having labels for issues (or using issue templates that auto-add labels to new issues) would be beneficial to searching existing issues. This idea would help prevent duplicates but requires (a separate issue and) a much more involved solution/proposal.

2bndy5:
resolves #415

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -89,11 +89,7 @@ env:
   NO_DEFAULT_FEATURES: ${{ inputs.NO_DEFAULT_FEATURES }}
   BUILD_DIR: ${{ github.workspace }}/built.d
 
-run-name: >-
-  Build Package
-  ${{ inputs.crate }}
-  ${{ inputs.version }} 
-  (${{ inputs.build_os }})
+run-name: Build Package ${{ inputs.crate }} ${{ inputs.version }} (${{ inputs.build_os }})
 
 jobs:
   build-popular-package:
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -89,6 +89,12 @@ env:
   NO_DEFAULT_FEATURES: ${{ inputs.NO_DEFAULT_FEATURES }}
   BUILD_DIR: ${{ github.workspace }}/built.d
 
+run-name: >-
+  Build Package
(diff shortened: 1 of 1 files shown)


2bndy5:
I created this from my phone, so I went with a mulit-line string that yml parsers concatenate into a single line. It should still have the same result as proposed in #415.

.github/workflows/build-package.yml
```diff
@@ -89,6 +89,12 @@ env:
   NO_DEFAULT_FEATURES: ${{ inputs.NO_DEFAULT_FEATURES }}
   BUILD_DIR: ${{ github.workspace }}/built.d
 
+run-name: >-
+  Build Package
+  ${{ inputs.crate }}
+  ${{ inputs.version }} 
+  (${{ inputs.build_os }})
```
2bndy5, line 96: I created this from my phone, so I went with a mulit-line string that yml parsers concatenate into a single line. It should still have the same result as proposed in #415.

Just in case this conflicts with style guidelines:
```suggestion
run-name: Build Package ${{ inputs.crate }} ${{ inputs.version }} (${{ inputs.build_os }})
```
Both are equivalent per YAML specifications.

2bndy5:
It seems like the some CI checks failed because my fork has no access to certain secret variables.
I think this could be mitigated with
```yml
if: ${{ secret.SOME_TOKEN }}
```
But I suspect it is more complicated because of the reusable workflows involved.

